PR #16, opened 2023-03-16
Howto debug golang operator via vscode

stuggi:


debugging.md
```diff
@@ -0,0 +1,52 @@
...
+* make sure [go extension for vscode](https://marketplace.visualstudio.com/items?itemName=golang.Go)
+is installed
+* install [delve debugger for go](https://github.com/go-delve/delve)
+```bash
+go install github.com/go-delve/delve/cmd/dlv@latest
+```
+* Deploy the needed operator CRDs/manifests either manually or install and uninstall via olm
+```bash
+oc login -u kubeadmin -p '12345678' https://api.ostest.test.metalkube.org:6443
+make install run
```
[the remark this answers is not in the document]
stuggi: yes, I had this first in a section call `how to run local` . I'll remove the run

debugging.md
```diff
@@ -0,0 +1,52 @@
...
+{
+        "version": "0.2.0",
+        "configurations": [
+        
+                {
+                        "name": "Launch Package",
+                        "type": "go",
+                        "request": "launch",
+                        "mode": "auto",
+                        "program": "<path to operator base dir>",
```
weinimo, line 34: Shouldn't that be:
`program": "<path to operator base dir>/main.go",`?

Also we should probably add the line
`            "cwd": "<path to operator base dir>",
`
or, what I personally use:
`            "cwd": "${fileDirname}",`

I am not sure if the relative path for OPERATOR_TEMPLATES below would work otherwise, if the .vscode project is not in the operator's root directory itself.

abays (approved):
/lgtm

debugging.md
weinimo, line 34: Maybe we could also add a task and add a line:
`            "preLaunchTask": "Run make",`
and the respective tasks.json. I use something like that:
`{
    "version": "2.0.0",
    "tasks": [
        {
            "label": "Run make",
            "type": "shell",
            "command": "make install manifests generate fmt vet",
            "group": {
                "kind": "build",
                "isDefault": true
            },
            "presentation": {
                "reveal": "always",
                "panel": "new"
            },
            "options": {
                "cwd": "<path to the operator base dir>",
                "env": {
                    "OPERATOR_TEMPLATES": "./templates",
                    "PATH": "~/.crc/bin/oc:${env:PATH}"
                }
            },
            "problemMatcher": []
        }
    ]
  }
`

weinimo (commented):
Thank you for working on documentation. I appreciate it. :+1:

debugging.md
stuggi: > Shouldn't that be: `program": "<path to operator base dir>/main.go",`?
> 
referencing the main package folder works for me as mentioned in https://vscode-debug-specs.github.io/go/ , `program must be main package folder or *.go file`
But seems the information is different on different sites. https://code.visualstudio.com/docs/editor/debugging has `executable or file to run when launching the debugger`

> Also we should probably add the line ` "cwd": "<path to operator base dir>",` or, what I personally use: ` "cwd": "${fileDirname}",`
> 
> I am not sure if the relative path for OPERATOR_TEMPLATES below would work otherwise, if the .vscode project is not in the operator's root directory itself.

good point, will add it

stuggi: thats a good idea, will add it

stuggi: updated . its now generic that you won't have to specify the paths

PR #20, opened 2023-03-29
Propose design test strategy

gibizer:


stuggi (commented):
+1 from my side, thanks for writing this @gibizer !

test_strategy.md
```diff
@@ -0,0 +1,57 @@
+# Test strategy
+
+## Desing test
```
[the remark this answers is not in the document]
gibizer: Now I run a spellcheck on the doc. :)

test_strategy.md
```diff
@@ -0,0 +1,57 @@
...
+   Each operator can extend the list of linters but we need a common minimum
+   set across each operators.
+
+2. Unit tests covering functions or types using go’s standard
+   [testing](https://go.dev/doc/tutorial/add-a-test) lib and
+   [gomega](https://onsi.github.io/gomega/)
+
+3. Component tests (also called functional or service tests) covering the
+   reconciliation of a single CRD or a set of CRDs from the same operator
+   using [EnvTest](https://book.kubebuilder.io/reference/envtest.html)
```
dprince, line 26: I'm -2 on having the above tests expected at this point. I think it is fine for some groups to use these tests, but I don't yet want to be called out on having to write these. Of course there is some value in any tests... but it seems to me the investment in the EnvTest's in particular is quite high.

test_strategy.md
```diff
@@ -0,0 +1,57 @@
+# Test strategy
+
+## Design test
+These are the checks and tests expected to be provided and maintained by the
```
dprince, line 4: perhaps "expected to be provided" is a bit strong at this point.

dprince: how about "the tests below have proven to be useful and are recommended for developing operators"

gibizer: I cannot feel confident to propose non trivial patches to operators not having this test coverage. I do believe not having this test coverage also slow other developers down providing those patches.

test_strategy.md
gibizer: The goal of this PR is not to guild anybody. But to agree on a common set of rules that, I think, would make both the product and the developer experience better.

dprince:
I am for the record not against folks writing tests and do support the general effort. If you are doing TDD and that helps your daily workflow go for it. I do think there are parts of our k8s-operators that are still largely in a POC state (with no unit/functional test coverage). It seemed the intent of this document was to say well, stop work and immediately add test coverage across the board. Perhaps it breaks some rules and design principles to say so but I would advocate that in these areas we might benefit from exploring the performance and user interfaces a bit more first. Like are we on the right path at a high level (especially on the dataplane side perhaps). Before adding unit/functional tests lets vet, or "shake out" these design concerns first... and that doing so might actually save us time. Maybe what I'm actually arguing for is another category of performance tests should come first in some cases (with lightweight functional/unit) tests peppered in where needed. Essentially a "test lite" approach for POC that allows us to *fail faster* in our interfaces design with regards to how they impact performance and the ultimate user interfaces we are exposing.

dprince:
> I am for the record not against folks writing tests and do support the general effort. If you are doing TDD and that helps your daily workflow go for it. I do think there are parts of our k8s-operators that are still largely in a POC state (with no unit/functional test coverage). It seemed the intent of this document was to say well, stop work and immediately add test coverage across the board. Perhaps it breaks some rules and design principles to say so but I would advocate that in these areas we might benefit from exploring the performance and user interfaces a bit more first. Like are we on the right path at a high level (especially on the dataplane side perhaps). Before adding unit/functional tests lets vet, or "shake out" these design concerns first... and that doing so might actually save us time. Maybe what I'm actually arguing for is another category of performance tests should come first in some cases (with lightweight functional/unit) tests peppered in where needed. Essentially a "test lite" approach for POC that allows us to _fail faster_ in our interfaces design with regards to how they impact performance and the ultimate user interfaces we are exposing.

pinikomarov:
+1, I think stability can increase feature development speed

test_strategy.md
slawqo: I think that what @dprince wrote makes sense for me. We shouldn't be too strong with wording and let teams to develop testing strategy in own way/time/etc.
I understand this document as something what is/should be final goal for all teams but it should give teams more freedom when they will implement that.

gibizer:
Marked it as draft now.
I will rephrase this to allow the developers of individual operators to pledge on a per operator basis for the strategy.

test_strategy.md
karelyatin: Thanks @gibizer for proposing it. Personally i found these tests useful as they provide faster responses and agree with the points already mentioned by You and Kamil.
I added these to some network operators already but not everyone in our DFG are familiar with it yet, same seems be true for other Teams too. As others have already planned work items for coming quarters expecting everyone to add these may not be right but can suggest and share the benefits of these so they can come to it whenever they think it's right time for them(atleast creating a Jira backlog would help teams so it's not forgotten fully).

fmount:
In general I agree w/ the proposal, `envTest` is a good framework that helps in the operators' development and avoid "hidden"  errors when a feature is added.
In storage an ongoing effort exists and tests are going to be added in both `Glance` and `Manila`, however, the idea is to be flexible enough to let the team decide when there's enough capacity to implement them (the plan is to start with the basic structure and plan to extend the coverage).
Given that I think we all agree on the topic, changing the wording mentioned above should be enough to move forward with this proposal.

gibizer:
I changed the blanket statement causing the push-back to:
> These are the checks and tests **encouraged** to be provided and maintained by the
developers of a given operator 

And I added a repo specific section where stronger statements can be made about specific repositories.

test_strategy.md
```diff
@@ -0,0 +1,37 @@
...
+4. Integration test using [kuttl](./kuttl_tests.md) on OCP
+   * Covering a single operator and its dependencies (e.g. placement-operator
+   tests will depend on keystone-operator and mariadb-operator)
+   * Covering our integrated set of operators via openstack-operator
+
+## Repository specific rules
+
+### nova-operator
+* Each PR changing implementation logic have to include relevant EnvTest
+coverage
```
gibizer, line 37: @SeanMooney @mrKisaoLamb @bogdando @kk7ds Please express if you agree with this nova-operator specific hard rule.

mrKisaoLamb: ++, strongly agree

test_strategy.md
```diff
@@ -0,0 +1,37 @@
...
+3. Component tests (also called functional or service tests) covering the
+   reconciliation of a single CRD or a set of CRDs from the same operator
+   using [EnvTest](./envtest.md)
+
+4. Integration test using [kuttl](./kuttl_tests.md) on OCP
+   * Covering a single operator and its dependencies (e.g. placement-operator
+   tests will depend on keystone-operator and mariadb-operator)
+   * Covering our integrated set of operators via openstack-operator
+
+## Repository specific rules
```
stuggi, line 33: entest coverage is also part of networking CRD in infra-operator. we should proceed and have coverage for changes/features added to infra-operator networking. for other components there we add them at some point.

stuggi: also lets start have test coverage in lib-common when adding new stuff there. where lib-common logic is not directly calling k8s APIs,  unit test coverage is required

gibizer: Thanks for the suggestions. I proposed a follow up in https://github.com/openstack-k8s-operators/docs/pull/30 to document these. Please check that I understood the rule from your comments properly.

test_strategy.md
bogdando: LGTM

dprince:
@gibizer thanks for accommodating the language update here. And for pressing through with this. Well done

PR #22, opened 2023-04-11
Operator controller-manager pod labels doc

abays:
Document describing how we approach operator `controller-manager` pod labeling such that it suits both aggregate and unique selection purposes.


stuggi (approved):
/lgtm

PR #24, opened 2023-05-09
Explain why we decided to use webhooks

abays:
Amend the webhooks doc to explain why we have introduced webhooks in the first place.


stuggi (approved):
/lgtm

PR #25, opened 2023-05-10
More use case examples for default images via env vars and webhooks

abays:
Further fleshing-out the justification for default OpenStack service container images via environment variables and webhooks.


olliewalsh (approved):
/lgtm

PR #31, opened 2023-05-23
Propose Service Bootstrap config design

fmount:
The patch represents a proposal to find a common pattern for services bootstrap.

### DIFF
--- a/service_config.md
+++ b/service_config.md
@@ -1,0 +1,227 @@
+#  Services configuration / k8s operators
+
+## Goal
+
+Describe an approach that can be adopted across the operators to improve the
+deployment aspect and the way secrets and config files are generated. The
+solution proposed in this document, in the first place, is to give an operator
+the ability to render and inject sensitive snippets within the service config
+deployed in the OpenStack controlplane.
+
+### Note:
+
+The document is not supposed to cover the implementation details that might
+converge or diverge according to a given operator, the number of deployed
+services, and other potential requirements that are not the same across the
+board.
+
+
+## Proposed approach
+
+The use cases are concentrated in two main aspects:
+
+* Provide a common interface that can be used to build additional config
+  snippets containing sensitive information using k8s `Secrets` instead of
+  `ConfigMaps`
+
+* Provide a common pattern to build regular service config:
+  * they will be processed via golang templates and mounted in a Pod from a
+    `ConfigMap` (which is what is currently happening)
+  * part of them will be rendered in a `Secret` and avoid exposing sensitive
+    information related to the deployment aspect
+
+
+## Basic deployment template
+
+Usually, the basic deployment config information is rendered in a `ConfigMap`.
+However, the idea is to store sensitive information in a `Secret` that will be
+mounted in the service Pod. For this reason, a given operator should be
+responsible to process “on board” the information coming from the main
+`osp_secret`, and generate a new `Secret` that matches to the deployment golang
+template that will be used by other service components.
+
+```
+                                                    +------------------------------------------------------------------------------------+
+1. The template is processed by the operator        | [database]                                                                         |
+2. A new Secret is created and  mounted as a Volume | connection = mysql+pymysql://{{.DBUser}}:{{.DBPassword}}@{{.DBHost}}/{{.Database}} |
+   by the Pod                                       |                                                                                    |
+3. Kolla sync src / dest and applies the expected   | [keystone_authtoken]                                                               |
+   ownership/permissions                            | password = {{.Password}}                                                           |
+                                                    | ...                                                                                |
+                     +----------------------------- | ...                                             (1)                                |
+                     |                              | [service_user]                                                                     |
+                     |                              | password = {{.Password}}                                                           |
+                     |                              +------------------------------------------------------------------------------------+
+                     |
++----------------------------------------+      +----------------------------------------------------------------------------------------+
+| apiVersion: v1                   (2)   |      |  “config_files”: [        (3)                                                          |
+| metadata:                              |      |                  ...                                                                   |
+|   name: 01-<service>-deployment        |      |      {                                                                                 |
+|   namespace: openstack                 | ===> |        "source": "/var/lib/config-data/deployment/01-<service>.conf",|
+| stringData:                            |      |        "dest": "/etc/<service>/<service>.conf.d/01-<service>.conf",  |
+|   01-<service>.conf: |                 |      |        "owner": "<service>",                                                           |
+|        <data>                          |      |        "perm": "0600"                                                                  |
++----------------------------------------+      |      },                                                                                |
+                                                +----------------------------------------------------------------------------------------+
+```
+
+As the diagram above depicts, a given operator is supposed to implement the
+logic to build and reconcile the `Secret` when the template has been processed.
+The `Secret`, mounted to the resulting service Pod, in the very last step of
+this process can be synced by kolla to the destination directory: instead of
+having a bash script (e.g. `init.sh`) doing the `chown` on the destination folder,
+we'll rely (where possible) on kolla that allows setting ownership and
+permissions, as well as copying optional files in case additional configuration
+is provided.
+
+## Render multiple secrets: “CustomServiceConfigSecrets” interface
+
+Currently the `OpenStack` storage operators expose the
+`customConfigServiceSecrets` parameter, which provides a mechanism for
+specifying service configs via Secrets. Instead of specifying sensitive config
+data directly in the `customServiceConfig`, a cloud admin can place sensitive
+data in a `Secret`, and reference the secret by name in the service's
+`customServiceConfigSecrets` as a list of `Secret` names, that will be
+iterated without any form of sorting (the same order how they appear in the list
+is used).
+
+```go
+for _, name := range instance.Spec.CustomServiceConfigSecrets {
+  confSecret, _, err := oko_secret.GetSecret(ctx, helper, name, instance.Namespace)
+    if err != nil {
+      // Secret not found or unable to retrieve the secret, returning err
+      return ctrlResult, err
+    }
+    confSecrets = append(confSecrets, *confSecret)
+}
+```
+
+Such parameter is used to select and collect existing k8s Secrets (without
+decoding any value in plaintext) and provide them as Volumes/VolumeMounts, that
+are processed by the operator and mounted in the target Pod.
+
+```
+     +----------------------------+
+     | customServiceConfigSecret: |
+ +---|   - service-secret1        |--------------------
+ |   |   - service-secret2        |                   |
+ |   +----------------------------+                   |
+ |         |                                   +-------------+
+ |         |                                   | - snippet 1 |
+ |  +------------------------------------+     | - snippet 2 | 
+ |  | apiVersion: v1                     |     | - snippet 3 |
+ |  | kind: Secret                       |     +-------------+
+ |  | metadata:                          |            |
+ |  |   name: service-secret1            |            |
+ |  |   namespace: openstack             |            |
+ |  | stringData:                        |  +----------------------------------+
+ |  |   snippet1: |                      |  | 04-secret-<service>.conf         |
+ |  |   [logger_root]                    |  |                                  |
+ |  |   level=WARNING                    |  | [logger_root]                    |
+ |  |   handlers=stdout                  |  | level=WARNING                    |
+ |  |   snippet2: |                      |  | handlers=stdout                  |
+ |  |   ##################               |  | ##################               |
+ |  |   # Log Formatters #               |  | # Log Formatters #               |
+ |  |   ##################               |  | ##################               |
+ |  |   [formatter_normal]               |  | [formatter_normal]               |
+ |  |   format=(%(name)s): s%(message)s  |  | format=(%(name)s): s%(message)s  |
+ |  +------------------------------------+  | [formatters]                     |
+ |                                          | keys=normal                      |
++-------------------------+                 +----------------------------------+
+| apiVersion: v1          |                           |
+| kind: Secret            |                           |
+| metadata:               |                           |
+|   name: service-secret2 |                           |
+|   namespace: openstack  |        1. Mounted as Volume by Service Pod
+| stringData:             |
+|   snippet3: |           |        2. `kolla_set_configs && kolla_start`: sync
+|  [formatters]           |           the resulting secret as done for the other
+|  keys=normal            |           regular config files
++-------------------------+
+```
+
+An example of the `customServiceConfigSecret` usage can be found in Manila,
+where this parameter has been used to test the
+[NetApp backend](https://gist.github.com/gouthampacha/1b5681104ee066b5bd2c702b29376199)
+
+### Note
+
+No one prevents the cloud admin to hold multiple Secrets where each secret can
+have many (even overlapping) oslo keys or config snippets. As stated earlier,
+secrets are processed in the order they appear in the CR parameter, and the
+concatenation of secrets containing duplicated sections can be problematic and
+presents constraints, hence it's strongly recommended to provide the same
+snippet or key only once. If multiple snippets are provided in a single Secret
+referenced in `customServiceConfigSecrets`, a predictable iteration order
+should be provided, hence the snippets are applied to the service config in the
+lexicographic ordering of the keys of those snippets.
+
+As a result of this strategy, the service presents a layout similar to the
+following:
+
+```
+00-default.conf            => default configs generated by operator. This is stored in ConfigMap or Secret
+01-deployment.conf         => default configs generated by operator, which contains credentials such as [database] connection.
+                              This is stored in Secret
+02-global.conf             => custom configs provided by users via top-level customServiceConfig. This is stored in ConfigMap or Secert
+03-service.conf            => custom configs provided by users via service level customServiceConfig. This is stored in ConfigMap or Secret
+04-secrets.conf            => custom configs provided by users via service level customServiceConfigSecrets, which contains credentials.
+                              This is stored in Secret and would not be present if no secrets are provided
+```
+
+
+The service will pass the `--config-dir` parameter to point to the
+`<service>.conf.d` directory where all the config files listed above are
+rendered.
+
+```
+"command": "/usr/bin/<service> --config-dir /etc/<service>/<service>.conf.d"
+```
+
+If this strategy is not available, the `init container` that executes a start
+script (e.g., `init.sh`) won't be removed, and the logic that generates the
+layout mentioned above will be implemented in the init script until the missing
+support is added in the upstream project.
+Init containers are still required in some cases: for instance, if a given
+service needs to be exposed via Multus on a particular network and the the IP
+information is required for its config (and the mentioned IP information is
+only available after the Pod is created and started), using an init container
+will help addressing such scenario.
+
+## Conclusion
+
+The model described here allows to reach many goals:
+- when possible, if a given service has no particular requirements, remove the
+  `initContainer` and the related scripts that are no longer required to start
+  the service: the `deployment Secret` is generated by the operator according
+  to the parameters defined in the service CR and the data retrieved by the
+  initial `osp_secret`
+
+- `Kolla` is still used  to copy files from `src` to `dst`, as they are
+  rendered and mounted accordingly with the right permissions in the
+  destination directory (*)
+
+- Operators’ controllers are able to parse many secrets referenced by the
+  `CustomServiceConfigSecrets` parameter and merge them into a **single**
+  `Secret` which is passed to the deployment and mounted to the `Pod` in the
+  target directory (**)
+
+Due the reasons mentioned above, `kolla` is still the target tool used to start
+the service.
+
+(\*) Mounting `Secrets` in the same destination directory where the `Configmap`
+files are synced currently generates permission related issues, and passing the
+`SubPath` to the `VolumeMount` doesn’t solve the problem. Removing Kolla from
+the picture doesn’t add much value rather than keeping it
+
+(\*\*) 04-<service>.conf is generated by the operator, and the `data` field is
+nothing more than the concatenation of the data retrieved by the list of the
+secrets specified in the service CR
+
+
+## Resources
+
+The work described in this document is supported by the `Glance` patch that has
+been tested via the `meta-operator` driven deployment:
+
+* [Glance PoC](https://github.com/openstack-k8s-operators/glance-operator/pull/221)
--- a/README.md
+++ b/README.md
@@ -17,3 +17,4 @@ Documentation related to podified controlplane.
 * [Strategy to provide files to services](extra_mounts.md)
 * [Ceph](ceph.md)
 * [Hyperconverged Infrastructure](hci.md)
+* [Service config and bootstrap](service_config.md)


service_config.md
```diff
@@ -0,0 +1,194 @@
...
+
+
+## Render multiple secrets: “CustomServiceConfigSecret” interface
+
+Currently the `OpenStack` storage operators expose the
+`customConfigServiceSecret` parameter, which provides a mechanism for
+specifying service configs via Secrets. Instead of specifying sensitive config
+data directly in the `customServiceConfig`, a cloud admin can place sensitive
+data in a `Secret`, and reference the secret by name in the service's
+`customServiceConfigSecrets` as a list of `Secret` names. Such parameter is
```
abays, line 85: This section references both `customConfigServiceSecret` and the plural version `customConfigServiceSecrets`.  Then below on L 92 it mentions the singular `customConfigServiceSecret` again.  Just trying to understand which one it should actually be, or is this use of both singular and plural intentional?

konan-abhi: I think it should be 'customServiceConfigSecret'

service_config.md
```diff
@@ -0,0 +1,194 @@
...
+the ability to render and inject sensitive snippets within the service config.
+
+### Note:
+
+The document is not supposed to cover the implementation details that might
+converge or diverge according to a given operator, the number of deployed
+services, and other potential requirements that are not the same across the
+board.
+
+
```
gibizer, line 17: Does this proposal targets service config for services running on the EDPM nodes or only targeting services running in k8s pods?

service_config.md
```diff
@@ -0,0 +1,194 @@
...
+    `ConfigMap` (which is what is currently happening)
+  * part of them will be rendered in a `Secret` and avoid exposing sensitive
+    information related to the deployment aspect
+
+
+## Basic deployment template
+
+Usually, the basic deployment config information is rendered in a `ConfigMap`.
+However, the idea is to store sensitive information in a `Secret` that, during
+the bootstrap of the service, will be mounted in the service Pod. For this
```
gibizer, line 37: Is it OK if an operator does not use init container to bootstrap a service pod but instead this secret will be mounted to the Pod for the lifetime of the Pod?

service_config.md
```diff
@@ -0,0 +1,194 @@
...
+is provided.
+
+
+## Render multiple secrets: “CustomServiceConfigSecret” interface
+
+Currently the `OpenStack` storage operators expose the
+`customConfigServiceSecret` parameter, which provides a mechanism for
+specifying service configs via Secrets. Instead of specifying sensitive config
+data directly in the `customServiceConfig`, a cloud admin can place sensitive
+data in a `Secret`, and reference the secret by name in the service's
```
gibizer, line 84: What will be the expected structure of the Secret? 
* a) directly storing key value pairs in the secret that are interpreted as oslo config options and their values? If so how we model oslo config sections
* b) a Secret expects to hold a single key value pair. Where the key will be fixed (like `snipet`) and the value is a string that holds a oslo config file snippet.
* c) a Secret can have many key value pairs. Each key is a name of a config file and the value is string that holds an solo config file snippet. How will the operator order the different files from a single Secret in the conf.d directory?

What will happen if the user specifies more than one secret name in `customServiceConfigSecrets`? Will the operator renders them as separate files. Therefore not one `04-secrets.conf` will be created but multiple files. If so I suggest to name the files based on `<index>-<name-of-the-secret>.conf` and use a running index that ensures the order of the secrets in `customServiceConfigSecrets` is the same as the order of the files in the conf.d dir.

service_config.md
```diff
@@ -0,0 +1,194 @@
...
+`<service>.conf.d` directory where all the config files listed above are
+rendered.
+
+```
+"command": "/usr/bin/<service> --config-dir /etc/<service>/<service>.conf.d"
+```
+
+If this strategy is not available, the `init container` that executes a start
+script (e.g., `init.sh`) won't be removed, and the logic that generates the
+layout mentioned above will be implemented in the init script.
```
gibizer, line 152: I suggest that in this case we raise an upstream blueprint / RFE to add support for it.

service_config.md
gibizer: Looking more at the example below I think I can answer some of my questions.

The `customServiceConfigSecrets` field is a list of string field, where each string is the name of a k8s Secret resource.

Each of these Secrets can have arbitrary amount of arbitrarily named a keys (e.g. snipet, foo, bar) the value at these keys are olso config file snippets. The the operator will concatenate all the values from all the secrets to get 04_secret.conf. E.g. something like:
```golang
var config4 string = ""
for _, secret := range instance.Spec.CustomServiceConfigSecrets {
    for _, v := range secret.Data {
        config4 += v + "\n"
    }
}
```

The oslo config allows repeating a configuration key. repetition has different meaning depending on the type of the oslo config option. The repetition can mean override the previous value, or can mean add the new value to the list of previous values of the config option. Due to the former case we need to be very explicit about the order of the concatenation as the last value of the repeated config option will be used by the service. In the above pseudo-code the secretes are iterated in the order of the list in `customServiceConfigSecrets` so that can be documented to the end user. But the iteration of the secret.Data keys are undefined. (golang map iteration order is random). Even if we sort the keys in the secret.Data we need to be very explicit in the documentation that even though we are not use the keys from the Secret directly, renaming a key could result in a different iteration order and therefore a different service config.

I suggest to be more explicit about the order and use the keys from the secret.Data as the name of the files we actually produce for the service. Then we can point to the oslo config doc that states the order in which the files are loaded. (This means instead of 04_secret.conf we might generate arbitrary amount of fiels in conf.d)

service_config.md
```diff
@@ -0,0 +1,194 @@
...
+As a result of this strategy, the service presents a layout similar to the
+following:
+
+```
+00-default.conf            => default configs generated by operator. This is stored in ConfigMap
+01-deployment-secrets.conf => default configs generated by operator, which contains credentials such as [database] connection . This is stored in Secret
+02-global.conf             => custom configs provided by users via top-level customServiceConfig. This is stored in ConfigMap
+03-service.conf            => custom configs provided by users via service level customServiceConfig. This is stored in ConfigMap
+04-secrets.conf            => custom configs provided by users via service level customServiceConfigSecrets, which contains credentials. This is stored in Secret and would not be present if no secrets are provided
+```
```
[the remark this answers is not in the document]
gibizer: @SeanMooney made me realize that we don't want to collect configuration data that contains sensitive information. So actually merging the config data into a single CM / Secret would not help with that. So please ignore me here.

service_config.md
SeanMooney: i dont think its a case of is it ok.
that should be the default after this change is adopted.

we should not require any init containers for configs

gibizer: Then I think we should drop the bootstrap wording from this.

service_config.md
SeanMooney: all service that use oslo.cofnig supprot 

/etc/<service>/<service>.conf.d by default.

so i think this would only be relevant for non OpenStack services
like ovs.

we may or may not want to expose configurabltiy of those services.

for example we currently don't plan to expose configuring libvirt unless we have a customer usecase that forces that. currently we do not have a usecase that requires it.

i agree with gibi if an OpenStack service cannot support config directories this should be addressed upstream in that service.

service_config.md
```diff
@@ -0,0 +1,194 @@
...
+  data retrieved by the initial `osp_secret`;
+
+- `Kolla` is still used  to copy files from `src` to `dst`, as they are
+  rendered and mounted accordingly with the right permissions in the
+  destination directory (*)
+
+- Operators’ controllers are able to parse many secrets referenced by the
+  `CustomServiceConfigSecrets` parameter and merge them into a **single**
+  `Secret` which is passed to the deployment and mounted to the `Pod` in the
+  target directory (**)
```
SeanMooney, line 174: I'm not convinced we need to support refining multiple secrets via
CustomServiceConfigSecrets

this seems like an overcomplication to me.

what is the motivation for this functionality instead of just allowing a single secret to be used?

service_config.md
SeanMooney: i was orignally expectign the customServiceConfigSecret to be a single secret name with a single hardcoded key containing a string which is the content of the file.

service_config.md
fmount: sorry, my fault there, `customServiceConfigSecrets` (plural) is an array of secrets: they can be referenced by `name` in the CR and the result is that `04-glance-secrets.conf` that has, as a result, the concatenation of the secret Data.
```
for _, name := range instance.Spec.CustomServiceConfigSecrets {
		confSecret, _, err := oko_secret.GetSecret(ctx, helper, name, instance.Namespace)
		if err != nil {
			// Secret not found or unable to retrieve the secret, returning err
			return ctrlResult, err
		}
		confSecrets = append(confSecrets, *confSecret)
	}
```

I'll patch the plural in the next iteration.

service_config.md
fmount: No, the proposal is for services in the ctlplane: I think we can follow up and expand to the EDPM world as long as we're in agreement on this side.

service_config.md
fmount: Not for config, but there are services with different use cases where can just drop the init container: anyway, I'm ok about dropping the bootstrap wording.

service_config.md
fmount: `CustomServiceConfigSecrets` is an array of secrets referenced by name in the CR. 
They are merged (see my comment above) and result in a single secret mounted to the pod, but no one prevent you to use a single entry of the array.
Is this a limitation for you?

```golang
for _, secret := range secrets {
  // merge all the content coming from the referenced secret
  data := secret.Data
  for _, v := range data {
     mergedData = append(mergedData, v...)
  }
  // Define the new secret
  s := corev1.Secret{
  ObjectMeta: metav1.ObjectMeta{
	Name:      secretName,
	Namespace: instance.Namespace,
  },
  Data: map[string][]byte{
	secretName + ".conf": mergedData,
   },
}
```

service_config.md
fmount: I agree but I don't want to enforce services to implement this spec.
I think it's great that we can come up with something we share and agree on, but we can have exceptions, it shouldn't be a problem given that the limitation is in the service itself (not supporting --config-dir for instance).
A given service can take this, follow up upstream with the relevant people and team, and come back to this solution.
How does it sound?

service_config.md
SeanMooney: how are they mounted int eh pod? where are we specifying how to generate the kolla config json to copy them to there final location and set the permissions?

fmount: yeah, the basic idea is to keep kolla doing the copy of the resulting (and predictable) optional file called `04-secrets.conf` (or whatever will be the name).
This means that in `pkg/<service>/volume.go` we can think about expanding `Volumes` and `VolumeMounts` arrays and mount the secrets we have (for secrets <- plural I mean `01-deployment.conf` and `04-secrets.conf` (regarding the naming, given it's internal, I'm also ok to keep something like `01-glance.conf`, `02-glance.conf`, `03-glance.conf`, `04-glance.conf` and define them as `const` in the operator, so we have a predictable amount of things we are copying).
e.g.

```golang
func GetVolumes(
	name string,
	pvcName string,
	extraVol []glancev1.GlanceExtraVolMounts,
	svc []storage.PropagationType,
	confSecret []string,
) []corev1.Volume {
    volumes := []corev1.Volume{
    ...
    ...
    ...
    }
    if len(confSecret) > 0 {
        for _, secret := range confSecret {
            sec, _ := GetConfigSecretVolumes(secret)
            volumes = append(volumes, sec)
         }
     }
    return volumes
}
``` 

where `GetConfigSecretVolumes` is a generic function that returns (corev1.Volume, corev1.VolumeMounts) associated to a generic secrets.
By doing this I can pass all the secrets generated by the operator and get them mounted to the Pod.
Kolla should be able to do the rest (it's mostly useful to adjust permissions and copy files to the `--config-dir` destination, given that mounting multiple (heterogeneous) sources in the same mountpoint doesn't work.

SeanMooney: i see i still dont see a stong usecase for allowing this to be an array but yes i was expecting CustomServiceConfigSecret to be the name of a single secret with a key
```
kind:secret
stringdata:
    configData: |
      [netapp]
      user=net
      password=app

```

service_config.md
gibizer: Sure. Let's state that this proposal only targets services running in k8s pods.

SeanMooney: part of the inital feedback i provided is that we should not try and split this along those lines since there is disagreement on what is the control plane.
just because its on an edpm host does not mean its not part fo the control plane.

with that said if other service s want have different config generation for depm nodes that is there choice but it would be explicitly against one of our early design goals which was to have no config generation in ansible.

for the nova operator anything we implement for openshift will also be used for the edpm nodes
we already excplivly generate service configuration for nova-comptue and libvirt using golang templates.

ovn/neutron however are currently split in how they are configured
hopefully they will see value in converging on this. proposal for there edpm configs too.

service_config.md
gibizer: 1. This is also potentially overriding keys between secrets.
```
mergedData = append(mergedData, v...)
```
E.g. if there are two secretes in `customServiceConfigSecrets` both having a single key `foo` with different snippets as values then this code will simply drop the value of the `foo` key from the first secret as the second will override it.

2. Also this code does not concatenate snippets within the resulting single secret into a single key that can be mapped to 04_secret.conf.

gibizer: Sorry disregard my comment. I missunderstood the snipet.

gibizer: So my only remaining comment here is that the concatenation order needs to be well defined as the meaning of resulting config file depends on it. The ` for _, v := range data {}` construct will iterate the map in a random order which is problematic.

fmount: Overriding: yes -> as I mentioned, it blindly concatenate things: the secrets are created by the human operator, not sure we have too much control over that content;
Concatenate: it works because I tried it, but we can do the `strings()` version +"\n", I'm not against that approach

service_config.md
gibizer: Sounds goo.

service_config.md
gibizer: > Overriding: yes -> as I mentioned, it blindly concatenate things: the secrets are created by the human operator, not sure we have too much control over that content;

I see two major option to make this clear for the human operator:
* a) remove ambiguity by not allowing more than one config snippet as input for 04-secrets.conf. This would mean the `customServiceConfigSecret` needs to hold only a single Secret, with a single (probably predefined) keys. Then no concatenation happens so no ordering issue araises.

* b) define a static ordering and document it clearly. `customServiceConfigSecrets` can hold multiple secrets and each secret can have many, (even overlapping) keys. We document that we concatenate the content of the secrets in `customServiceConfigSecrets` by
  * iterating the list of Secrets in the order how they appear the list in `customServiceConfigSecrets`
  * iterating the keys in each Secret in lexicographical order

E.g.:
```
customServiceConfigSecrets:
    foo-secret
    bar-secret
//--
FooSecret
    data:
        pear: |
            an-oslo-conf-option = foo-pear
        apple: |
            an-oslo-conf-option = foo-apple
//--
BarSecret
    data:
        pear: |
            an-oslo-conf-option = bar-pear
        apple: |
            an-oslo-conf-option = bar-apple
```
Will result in the 04-secret.conf file:
```
            an-oslo-conf-option = foo-apple
            an-oslo-conf-option = foo-pear
            an-oslo-conf-option = bar-apple
            an-oslo-conf-option = bar-pear
```
So the lexicographically biggest key of the last Secret from `customServiceConfigSecrets` wins, if mutliple keys define the same oslo config option.

SeanMooney: option A is my preference by far.
customServiceConfigSecret: myCustomSecret
//--
myCustomSecret:
  StringData:
    config: |
      [section-1]
      key=value
      [meaning]
      life=42

service_config.md
stuggi: init containers are still required in some cases. e.g. there ovs-controller requires an init container to run ovs-ctl commands before the service container starts.
Also if services need to be exposed via multus to networks and the the IP information is required for a config, that would be also a use case for an init container for config. That information is only available at pod start, e.g. OvnEncapIP - https://github.com/openstack-k8s-operators/ovs-operator/pull/75 . if there are services which require setting such information in a config it would be required to do so via an init container

SeanMooney: there may still be use-cases for them but I'm not sure if this is a good example.
this could be done by the use of a kolla_extended_start script
in the ovs-vswichtd container instead of an init container.

https://github.com/openstack-k8s-operators/tcib/blob/main/container-images/kolla/base/start.sh#L12-L15

infact the entire init.sh that is currently there could likely be moved to use kolla_extended_start since that was the exact usecase it extis to enable.

allowing injection of additional operations to be done on start

there may still be other reasons to use an init container but i think the point I'm trying to advance is when considering what to use them for we should hesitate before using them for config file modification.

service_config.md
```diff
@@ -0,0 +1,194 @@
...
+
+An example of the `customServiceConfigSecret` usage can be found in Manila,
+where this parameter has been used to test the
+[NetApp backend](https://gist.github.com/gouthampacha/1b5681104ee066b5bd2c702b29376199)
+
+
+## Conclusion
+
+The model described here allows to reach many goals:
+- remove (entirely) the `initContainer` and the related scripts that are no
```
stuggi, line 162: as from my comment above there might be still a use case for init container, if e.g. IP information is required to setup the service/service-config.

service_config.md
stuggi: agree we should prevent using them to create service configs where possible. I just wanted to mention that init containers are not bad and there might be use cases as the one above. if kolla_extended_start is used or some init container I guess is up to the component?

SeanMooney: either works in both cases the script is intended to be injected into the container not baked in so there is little different.

kolla_extended_start exists because just using podman/docker at least when kolla was first started you could could not run init containers.

podman actually allow you to define pods not just container so they might have the concept of init contaienrs in ther model.

kolla_extended_start would work for both openshift an rhel nodes but functionally they serve the same usecase.

so the service should just use what make the most sense in a given context

gibizer: My original intention with my comment was simply that 

> [...] store sensitive information in a `Secret` that, during
the bootstrap of the service, will be mounted in the service Pod.

this implies for me that the Secrets are only mounted during bootstrap. But in reality we mount them for the lifetime of the Pod.

service_config.md
bogdando: right, geven the outlined oslo config specifics, and (deep?)merging complexity (nesting lists to add or replace?, keys ordering, etc.) please specify the case for having multiple secrets referenced , which cannot be covered by a single secret (aside of the 1.5M size limitation, iirc)?

Akrog: Personally I think we should just document how secrets are used here, which is that operators will drop everything within a secret without following an order, so human operators cannot use a single secret to store different things.

For example, they probably don't want to use a single secret to store ALL their cinder backend secrets, otherwise they will be present in all the different cinder-volume pods even if they are not used.

gibizer: @Akrog Thanks for the use case. I think this is a good example to support multiple Secrets. I still don't see a use case to allow multiple snippets per Secret.

Akrog: I don't think it will be used, but I may be missing some use case.

In my opinion the reason to support multiple snippets per secret is simplicity.  Basically we say that we drop everything within a secret instead of explaining that we'll only use the first snippet.

bogdando: for multiple backends, ther could be still a single secret like:
```
kind: Secret
  metadata:
  name: cinder-volume-backends
data:
  backend1.conf: |
  backend2.conf: |
```

service_config.md
```diff
@@ -0,0 +1,194 @@
...
+|   01-deployment-<service>.conf: |      |      |        "owner": "<service>",                                                           |
+|        <data>                          |      |        "perm": "0600"                                                                  |
++----------------------------------------+      |      },                                                                                |
+                                                +----------------------------------------------------------------------------------------+
+```
+
+As the diagram above depicts, a given operator is supposed to implement the
+logic to build and reconcile the `Secret` when the template has been processed.
+The `Secret`, mounted to the resulting service Pod, in the very last step of
+this process is synced by kolla to the destination directory: instead of having
```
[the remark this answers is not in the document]
bogdando: directly mounted in /etc/cinder/cinder.conf.d hides all the original contents of that dir, exposing only files from the secret. That could become a problem for minor updates, for example, or gradual passwords rotation (where we might want to have both old config with the old password  alongside the updated config with the changed credentials). So we'' need to use subPaths

service_config.md
Akrog: If I understood correctly the current proposal is to only support whole secrets, not allow using specific keys.

In your example we wouldn't be able to mount "backend1.conf" and "fc_zone_secret", we would have to use everything.

The recommendation for cinder is going to be to have 1 pod per backend (there are many, many reasons why this is a superior option) and in that case mounting the whole secret would make services have more information that they should.  Don't get me wrong, it would work, but it may not be the best in terms of security.

If a customer prefers to go with the multi-backend per cinder-volume pod, then that configuration would be fine.

service_config.md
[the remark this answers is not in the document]
Akrog: What original contents of the directory?  The directory is empty and the defaults come from the operator.
How would you envision gradual passwords rotation working?  Because I think I'm missing something.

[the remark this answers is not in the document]
bogdando: @SeanMooney ^^ fyi. @Akrog please see [OSPRH-92](https://issues.redhat.com/browse/OSPRH-92) that explains the "gradual passwords rotation" w/o openstack services downtime. Does that also apply for Cinder backends access creds? If so, the backend config directory *might* need to maintain the "old" version of credentials, and the new (rotated) ones in another set of configs for some transitory time?

wrt "hidden" contents in cinder.conf.d: "As we mount a Kubernetes volume at a mount-point, and if we already have some content in it, then it would be hidden by our mount. But in some cases, we would not want to hide the existing content at the mount-point but want to add additional files/directories in parallel"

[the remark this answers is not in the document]
Akrog: > @SeanMooney ^^ fyi. @Akrog please see [OSPRH-92](https://issues.redhat.com/browse/OSPRH-92) that explains the "gradual passwords rotation" w/o openstack services downtime. Does that also apply for Cinder backends access creds? If so, the backend config directory _might_ need to maintain the "old" version of credentials, and the new (rotated) ones in another set of configs for some transitory time?

Sorry @bogdando, I still don't see how you envision the credentials rotation working...
Cinder services cannot use 2 credentials at the same time, and I doubt many OpenStack services can.
Once we decide to change credentials the services need to be restarted (which we can probably do progressively) with the new configuration.
So during the rotation we'll have services using the old credentials and some using the new credentials, and if both are valid then everything is fine.  At some point all services will be using the new credentials and the old credentials can be "deleted".

Am I mistaken in thinking that's how the rotation would work?


> wrt "hidden" contents in cinder.conf.d: "As we mount a Kubernetes volume at a mount-point, and if we already have some content in it, then it would be hidden by our mount. But in some cases, we would not want to hide the existing content at the mount-point but want to add additional files/directories in parallel". But if the directory is empty and all comes from the operator, that shouldn't be a problem

For cinder we don't currently use that together, and even if we were we don't want the RPM defaults, we want the defaults that are defined in our operator.  I thought we had all agreed that operators would have their own default configuration files stored in their repositories and would not use any other defaults, but I could be wrong here.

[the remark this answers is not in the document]
bogdando: yes, that makes sense

service_config.md
fmount: +1, I'm going to change the wording here, and as per your suggestion in the previous comment I'm going to state that there might be cases where we'd like to keep the init container to properly handle the service start (e.g. you need to know the IP address on the Multus network, which is known when the Pod is created and started).

service_config.md
fmount: I think this has been covered in another thread in this patch, where @Akrog pointed out the requirement for `cinder` of  having something like:

```
cinder-volume1:

    driver-config1-secret
    fc-zone-manager-secret

cinder-volume2:

    driver-config2-secret
    fc-zone-manager-secret
``` 

(driver credentials as well as shared FC zoning driver credentials).

service_config.md
fmount: > 



> > Overriding: yes -> as I mentioned, it blindly concatenate things: the secrets are created by the human operator, not sure we have too much control over that content;
> 
> I see two major option to make this clear for the human operator:
> 
>     * a) remove ambiguity by not allowing more than one config snippet as input for 04-secrets.conf. This would mean the `customServiceConfigSecret` needs to hold only a single Secret, with a single (probably predefined) key. Then no concatenation happens so no ordering issue araises.
> 
>     * b) define a static ordering and document it clearly. `customServiceConfigSecrets` can hold multiple secrets and each secret can have many, (even overlapping) keys. We document that we concatenate the content of the secrets in `customServiceConfigSecrets` by
>       
>       * iterating the list of Secrets in the order how they appear the list in `customServiceConfigSecrets`
>       * iterating the keys in each Secret in lexicographical order
> 
> 
> E.g.:
> 
> ```
> customServiceConfigSecrets:
>     foo-secret
>     bar-secret
> //--
> FooSecret
>     data:
>         pear: |
>             an-oslo-conf-option = foo-pear
>         apple: |
>             an-oslo-conf-option = foo-apple
> //--
> BarSecret
>     data:
>         pear: |
>             an-oslo-conf-option = bar-pear
>         apple: |
>             an-oslo-conf-option = bar-apple
> ```
> 
> Will result in the 04-secret.conf file:
> 
> ```
>             an-oslo-conf-option = foo-apple
>             an-oslo-conf-option = foo-pear
>             an-oslo-conf-option = bar-apple
>             an-oslo-conf-option = bar-pear
> ```
> 
> So the lexicographically biggest key of the last Secret from `customServiceConfigSecrets` wins, if mutliple keys define the same oslo config option.

Thanks @gibizer. We don't have control over the secrets created by the human operator, hence I'd like to document that  `customServiceConfigSecrets` can hold multiple secrets (because it's an array) and we strongly recommend to not store multiple snippets in a single secret because it can result in a not predictable configuration.
For the record, I think we could (and we should) do the iteration in lexicographical order (so that the last one win): it's something we had in `TripleO` with regard to the `Heat` parameter processing, but given this is a generic high level document we can leave this as a design choice for each operator and document it in the related section/ (e.g. see [1]).
Do you think it's reasonable?

[1] https://github.com/openstack-k8s-operators/glance-operator/blob/main/docs/dev/design-decisions.md

service_config.md
fmount: +1, I'll change the wording in the next iteration

service_config.md
fmount: Done

service_config.md
fmount: I think we can resolve this comment as long as it's clear other projects (e.g. cinder) have the needs to provide an array of secrets. @SeanMooney feel free to reopen in case you have additional comments, I'm just trying to clean a bit the patch and see if we have outstanding things to cover/address

service_config.md
gibizer: If it helps settle this sooner then I agree to scope it to services in k8s pods only. While I agree with @SeanMooney that the same logic described here should be applied to services running on EDPM nodes, I prefer settle this with a smaller scope sooner and extending it later, than holding the whole thing up due to the EDPM architectural discussion.

service_config.md
gibizer: > customServiceConfigSecrets can hold multiple secrets

OK by me as there is a use case for it mentioned above

> we strongly recommend to not store multiple snippets in a single secret

OK. 

> we could (and we should) do the iteration [of the keys in a Secret] in lexicographical order

*[clarification is from me]* OK

> but given this is a generic high level document we can leave this as a design choice for each operator and document it in the related section

I disagree here. I think the exact meaning (i.e. the resulting oslo configuration) of the multi keys Secret case should be defined here.  We don't want two operators defining the same `customServiceConfigSecrets` field with the same syntaxt **but with different semantic**. As that would be highly confusing and error prone to operator by the end user.

service_config.md
fmount: +1 I think we're on the same page here.

service_config.md
fmount: Done.

service_config.md
fmount: Yes, but I'm not sure we're defining how the operator should parse the list of secrets referenced via the `customServiceConfigSecrets` interface: I'm ok to propose to iterate in a given order (lexicographical) to avoid problems like the one described in the previous comments, I'm less sure about opening a `semantic` related conversation in this document.

fmount:
@gibizer @abays @stuggi except the last comment that I still discussing w/ gibi, do you have additional concern of we can approve this document at this point?
I know that nova is starting the implementation, same story for cinder and we have a WIP in glance. If you don't have additional concern we can spend some time on the implementation itself instead of high level design choices that at this point are pretty much sorted.

service_config.md
fmount: Done

gibizer (commented):
lgtm

abays (approved):
/lgtm

stuggi:
> @gibizer @abays @stuggi except the last comment that I still discussing w/ gibi, do you have additional concern of we can approve this document at this point? I know that nova is starting the implementation, same story for cinder and we have a WIP in glance. If you don't have additional concern we can spend some time on the implementation itself instead of high level design choices that at this point are pretty much sorted.

I think I am good, thanks. lgtm

fmount:
Looks like we have enough votes to move forward: I'm merging the patch at this point.

PR #32, opened 2023-05-24
[test-strategy]Add lib-common specific rules

gibizer:
This is based on the comments from @stuggi in
https://github.com/openstack-k8s-operators/docs/pull/20

### DIFF
--- a/test_strategy.md
+++ b/test_strategy.md
@@ -40,3 +40,13 @@ coverage
 * Each PR changing implementation logic targeting the
 [network](https://github.com/openstack-k8s-operators/infra-operator/tree/main/controllers/network)
 controllers have to include relevant EnvTest coverage.
+
+### lib-common
+* Adding new functionality needs
+  * either EnvTest coverage if the logic is calling k8s APIs. For example
+  [job](https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/test/functional/job_test.go)
+  * or unit test coverage if it isn't calling k8s APIs. For example
+  [conditions](https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/condition/funcs_test.go)
+
+  Except for the `test` and `test-operators` modules as they are test helpers
+  already.


gibizer:
I've added @fmount from storage and @slagle from edpm side as this affects the storage and ansible modules of lib-common too.

test_strategy.md
```diff
@@ -40,3 +40,13 @@ coverage
 * Each PR changing implementation logic targeting the
 [network](https://github.com/openstack-k8s-operators/infra-operator/tree/main/controllers/network)
 controllers have to include relevant EnvTest coverage.
+
+### lib-common
```
fmount, line 44: Ack and +1 on my side. 
FWIW when we started the ceph module at the time (just helpers, not much used in the operators ecosystem) we added something like [1], which seems to fit the purpose.
I'm pretty sure we can improve the `storage/` module and add test cases for the existing functions, but in general for functional tests I would defer to the operators implementing the feature to cover the `scenario` and double check (via `envTest`) the expected results.
@gibizer how does it sounds?

[1] https://github.com/fmount/lib-common/blob/extra_volumes/modules/storage/ceph/funcs_test.go

slagle:
+2 from me. We already have some initial unit test coverage in https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/ansible/inventory_test.go for the ansible related functionality.

test_strategy.md
gibizer: Cool I will update the patch. I looked through the storage module and I don't see any k8s API calls in there so the existing unit test feel OK to me as is. We can discuss any incoming new features in that module separately but I will assume for now that unit test coverage will suffice.

gibizer:
> +2 from me. We already have some initial unit test coverage in https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/ansible/inventory_test.go for the ansible related functionality.

Thanks. As far as I see the ansible module also does not reach out to k8s APIs so the existing unit test coverage is good as is and I will assume the same for the incoming new features for now as well.

test_strategy.md
gibizer: Ohh I already have the unit / functional distinction in the current docs patch. So I think both the storage and ansible module is belonging to the unit tested category.

PR #36, opened 2023-05-31
Document Ceph RBD Configuration

fultonj:
This document describes how to use CRs from the
openstack-k8s-operators project to configure OpenStack so that Glance, Cinder, and Nova use block storage from an external Ceph cluster. It does not require the use of install_yamls.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,3 +14,4 @@ Documentation related to podified controlplane.
 * [Operator controller manager pod labels](operator_controller_labels.md)
 * [Testing operators with EnvTest](envtest.md)
 * [Test strategy](test_strategy.md)
+* [Ceph](ceph.md)
--- a/ceph.md
+++ b/ceph.md
@@ -1,0 +1,256 @@
+# Ceph
+
+## Overview
+
+This document describes how to use CRs from the
+openstack-k8s-operators project to configure OpenStack
+so that Glance, Cinder, and Nova use block storage
+from an external Ceph cluster. It does not require
+the use of install_yamls.
+
+## Prerequisites
+
+- Access to a Ceph cluster
+- Access to an OpenShift cluster where the OpenStack operator is
+  already running
+- The reader knows how to create CRs from the openstack-k8s-operators
+  project to deploy OpenStack
+
+## Ensure Ceph is accessible via the Storage Network
+
+As per the [Networking document](networking.md)
+Ceph should be accessed via the storage network.
+The storage network is the same as Ceph's public_network. The Glance
+and Cinder pods should be able to access the storage network via
+MetalLB. The Nova compute containers should also be able to access the
+storage network. Glance, Cinder, and Nova will use a ceph.conf file
+containing the IPs of the Ceph monitors; these IPs should be within
+the storage network's IP range.  It is not necessary for OpenStack to
+access Ceph's cluster_network.
+
+## Create Ceph Pools for OpenStack
+
+The commands in this section should be run on a Ceph server.
+
+Create pools for Nova (vms), Cinder (volumes) and Glance (images).
+```
+for P in vms volumes images; do 
+  cephadm shell -- ceph osd pool create $P;
+  cephadm shell -- ceph osd pool application enable $P rbd;
+done
+```
+
+Create a cephx key which OpenStack can use to access the pools.
+```
+cephadm shell -- \
+   ceph auth add client.openstack \ 
+     mgr 'allow *' \
+	 mon 'allow r' \
+	 osd 'allow class-read object_prefix rbd_children, allow rwx pool=vms, allow rwx pool=volumes, allow rwx pool=images'
+```
+
+Export the cephx key and Ceph configuration file.
+```
+cephadm shell -- ceph auth get client.openstack > /etc/ceph/ceph.client.openstack.keyring
+cephadm shell -- ceph config generate-minimal-conf > /etc/ceph/ceph.conf
+```
+
+## Create a Ceph Secret
+
+Transfer the cephx key and Ceph configuration file from the previous
+section to a host which can create resources in the openstack
+namespace. Base64 encode these files and store them in two variables.
+```
+KEY=$(cat /etc/ceph/ceph.client.openstack.keyring | base64 -w 0)
+CONF=$(cat /etc/ceph/ceph.conf | base64 -w 0)
+```
+Use the variables to create a `ceph-conf-files` secret.
+```
+cat <<EOF > ceph_secret.yaml
+apiVersion: v1
+data:
+  ceph.client.openstack.keyring: $KEY
+  ceph.conf: $CONF
+kind: Secret
+metadata:
+  name: ceph-conf-files
+  namespace: openstack
+type: Opaque
+EOF
+
+oc create -f ceph_secret.yaml
+```
+
+The Ceph FSID can be extracted from the secret using the following
+command.
+```
+FSID=$(oc get secret ceph-conf-files -o json | jq -r '.data."ceph.conf"' | base64 -d | grep fsid | sed -e 's/fsid = //')
+```
+The above will be useful for configuration snippets covered later in
+the document.
+
+## Access the Ceph Secret via extraMounts
+
+Use
+[extraMounts](https://github.com/openstack-k8s-operators/lib-common/commit/c3752c8a43423c755adb16aebf644c304befc072)
+in CRs for pods which need to access the Ceph
+secret. For example, the sample
+[core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml](https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml)
+has the following in the `OpenStackControlPlane` CR so that the
+Glance and Cinder volume pods mount the cephx key and Ceph
+configuration files in /etc/ceph.
+
+```
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
+spec:
+  extraMounts:
+    - name: v1
+      region: r1
+      extraVol:
+        - propagation:
+          - CinderVolume
+          - GlanceAPI
+          extraVolType: Ceph
+          volumes:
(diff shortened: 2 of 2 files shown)


ceph.md
```diff
@@ -0,0 +1,245 @@
...
+            projected:
+              sources:
+              - secret:
+                  name: ceph-conf-files
+          mounts:
+          - name: ceph
+            mountPath: "/etc/ceph"
+            readOnly: true
+```
+The `OpenStackDataPlane` can also use `extraMounts`.
```
SeanMooney, line 126: this section feels a bit handwavy


im conficlitd between explaining just watch is required an providing a complete example.

perhaps we could provide examples of a complete
OpenStackControlPlane cr and OpenStackDataPlane/OpenStackDataPlaneRole/OpenStackDataPlaneNode crs

at the end

so we can discusee the snipits here in each section but then show what it looks like when its combined.

reition now it could be assume that each sech is used standalone as separate cr when infact these will be combined.

ceph.md
```diff
@@ -0,0 +1,245 @@
...
+```
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlane
+spec:
+  roles:
+    edpm-compute:
+	  nodeTemplate:
+        extraMounts:
```
gibizer, line 134: Is this an indentation error or just the github view got confused?

ceph.md
```diff
@@ -0,0 +1,245 @@
...
+          - name: ceph
+            mountPath: "/etc/ceph"
+            readOnly: true
+```
+When a CR containing the above is created, an Ansible pod
+running on OpenShift mounts the files in the Ceph secret
+and copies them to the EDPM host using the
+[edpm_ceph_client_files](https://github.com/openstack-k8s-operators/edpm-ansible/tree/main/roles/edpm_ceph_client_files)
+Ansible role. The Nova containers then have a copy of /etc/ceph on
+their that they can use to acccess the the cephx key and Ceph
```
gibizer, line 150: > copy of /etc/ceph on their that they can use

It feels like a word is missing somewhere here

gibizer (commented):
Thanks for documenting it I only have minor nits inline

ceph.md
fultonj: Yes, thanks. Fixed in the next PS.

ceph.md
```diff
@@ -0,0 +1,245 @@
...
+            [DEFAULT]
+            enabled_backends=ceph
+            [ceph]
+            volume_backend_name=ceph
+            volume_driver=cinder.volume.drivers.rbd.RBDDriver
+            rbd_ceph_conf=/etc/ceph/ceph.conf
+            rbd_user=openstack
+            rbd_pool=volumes
+            rbd_flatten_volume_from_snapshot=False
+            rbd_secret_uuid=_FSID_
```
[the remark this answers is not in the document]
fultonj: OK, I changed it to `$FSID`

ceph.md
fultonj: "The Nova containers then have a copy of /etc/ceph that they can use to acccess the cephx key and Ceph configuration file."

ceph.md
fultonj: Thanks for catching. Will be fixed in next PS.

Details: `nodeTemplate` had "<tab><space><space>" while all of the other strings in the yaml block had N spaces. My next revision will have 6 spaces in front of `nodeTemplate`.

ceph.md
fultonj: "The Nova containers then have a copy of /etc/ceph that they can use to acccess the cephx key and Ceph configuration file."

ceph.md
fultonj: > im conficlitd between explaining just [what] is required an providing a complete example.
> 
> perhaps we could provide examples of a complete OpenStackControlPlane cr and OpenStackDataPlane/OpenStackDataPlaneRole/OpenStackDataPlaneNode crs
> 
> at the end

I link to a complete example of `OpenStackControlPlane` node on lines 98, 158 and 185.
I can link to a complete example of `OpenStackDataPlane` too.

> so we can discusee the snipits here in each section but then show what it looks like when its combined.

Yes, I did want to have a single document which focussed on the necessary snippets required only for Ceph.

> reition now it could be assume that each sech is used standalone as separate cr when infact these will be combined.

I thought the including the `kind` of each CR it would be clear which CRs are separate and which could be combined. One of my pre-reqs on line 16 is that the reader knows to use the openstack-k8s-operators project and is looking to modify it to include Ceph. Thus, I focussed only on the diff but also provided an explanation.

gibizer (approved):
My nits were fixed so this looks good to me.

ceph.md
fultonj: I'll update the next PS to spell it out more and link to complete examples.

fmount: A link to the example is sufficient IMO, the whole `OpenstackControlPlane` CR is too verbose and should also be updated in case anything change, so I would stay on the short version, maybe with some "dots" in the middle, so the reader knows we're referring to something we can't just `oc apply -i`

fmount (approved):

ceph.md
```diff
@@ -0,0 +1,256 @@
...
+## Full Examples
+
+The examples above are focussed on showing how a
+single `OpenStackControlPlane` and `OpenStackDataPlane`
+CR can be modified to include Ceph configuration by adding
+`extraMounts` and `customServiceConfig`. Links to complete
+examples are below.
+
+- `OpenStackControlPlane`: [core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml](https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml)
+- `OpenStackDataPlane`: [dataplane_v1beta1_openstackdataplane_ceph.yaml](https://github.com/openstack-k8s-operators/dataplane-operator/blob/main/config/samples/dataplane_v1beta1_openstackdataplane_ceph.yaml)
```
fultonj, line 256: This link will work after this PR merges

https://github.com/openstack-k8s-operators/dataplane-operator/pull/216

ceph.md
fultonj: @SeanMooney I've added a section at the bottom called "Full Exampes" which I think points out what you are asking for. It links to the sample ControlPlane CR with Ceph config under the openstack-operator. The data-plane operator didn't have a ceph example so I created one in another PR and linked to it from the latest version of the document.

https://github.com/openstack-k8s-operators/dataplane-operator/pull/216

ceph.md
SeanMooney: ack

ceph.md
SeanMooney: thats no quite what i was askign for but it will do for now.

i was asking for you to embed it directly in the doc but i think long term we can have a better solution by maintaining a repo with example that are tested so linking for now i guess is ok.

thanks fro creating the datapalne-oeprator example

ceph.md
```diff
@@ -0,0 +1,256 @@
...
+            rbd_ceph_conf=/etc/ceph/ceph.conf
+            rbd_user=openstack
+            rbd_pool=volumes
+            rbd_flatten_volume_from_snapshot=False
+            rbd_secret_uuid=$FSID
+```
+The `$FSID` value above should contain the actual FSID as described
+in the "Create a Ceph Secret" section. The FSID itself does not need
+to be considered secret.
+
```
fmount, line 213: ## Configure Manila with native CephFS

Use a `customServiceConfig` to pass overrides to Manila's
configuration file. For example, the sample
[core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml](https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/samples/core_v1beta1_openstackcontrolplane_network_isolation_ceph.yaml)
has the following in the `OpenStackControlPlane` CR:

```
apiVersion: core.openstack.org/v1beta1
kind: OpenStackControlPlane
spec:
  extraMounts:
    ...
    manila:
        template:
           manilaShares:
             share1:
                customServiceConfig: |
                    [DEFAULT]
                    enabled_share_backends=cephfs
                    enabled_share_protocols=cephfs
                    [cephfs]
                    driver_handles_share_servers=False
                    share_backend_name=cephfs
                    share_driver=manila.share.drivers.cephfs.driver.CephFSDriver
                    cephfs_conf_path=/etc/ceph/ceph.conf
                    cephfs_auth_id=openstack
                    cephfs_cluster_name=ceph
                    cephfs_enable_snapshots=True
                    cephfs_ganesha_server_is_remote=False
                    cephfs_volume_mode=0755
                    cephfs_protocol_helper_type=CEPHFS
```

@fultonj Note: this will be available once [1] lands.

[1] https://github.com/openstack-k8s-operators/openstack-operator/pull/332

fmount (changes requested):
Can we add `Manila` to the picture while the patch is under review?

README.md
```diff
@@ -14,3 +14,4 @@ Documentation related to podified controlplane.
 * [Operator controller manager pod labels](operator_controller_labels.md)
```
fmount, line 14: .. and we can expand the title of the PR to cover CephFS other than rbd

fultonj: Can we please add manila in a follow up patch and not delay merging this one?

ceph.md
```diff
@@ -0,0 +1,256 @@
+# Ceph
+
+## Overview
+
+This document describes how to use CRs from the
+openstack-k8s-operators project to configure OpenStack
+so that Glance, Cinder, and Nova use block storage
```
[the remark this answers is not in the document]
fultonj: I prefer to have a follow up patch rather than scope creep this one.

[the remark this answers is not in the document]
fmount: Ack, so let's merge this one and we'll follow up for Manila, so we don't have to change the scope here.

fmount (approved):
Let's add Manila in a follow up patch

fultonj:
> Let's add Manila in a follow up patch

@fmount Thanks. I used your comments from this review to create this one on top of the now merged patch.

https://github.com/openstack-k8s-operators/docs/pull/37

PR #40, opened 2023-06-12
Update local webhooks procedure for new operator bundling

abays:
The local webhooks instructions are simplified now that we no longer bundle service operators within the OpenStack operator


stuggi (approved):
/lgtm

PR #41, opened 2023-06-13
[spec] Propose EDPM OpenStack service architecture

gibizer:


dprince:
@gibizer This came across my path today and I was a bit confused why it exists. We have at least 3 or 4 google docs discussing architectural options for Dataplane services. Is this meant to be a consolidation of the Google docs perhaps and you only want comments here now?

gibizer:
> @gibizer This came across my path today and I was a bit confused why it exists. We have at least 3 or 4 google docs discussing architectural options for Dataplane services. Is this meant to be a consolidation of the Google docs perhaps and you only want comments here now?

I'm trying to follow the pattern started in #31. Where after the initial discussion started to settle in the google docs the actual proposal is cleaned up and moved to this repo. This is intentionally in Draft state as I just finished writing this up as a cleanup of the google doc where the question / comments seems to be settled. 

I'm traveling this week. So this and #42 will remain in Draft state at least until next week. If you have comments feel free to provide it here or in the google docs (whichever is more convenient for you)  in the meantime. I will sync things up next week and we can also chat about the way forward.

spec/edpm-openstack-service-architecture.md
```diff
@@ -0,0 +1,323 @@
...
+
+Various OpenStack componentes have agents that need to run on the External
+DataPlane Node. For example OpenStack Nova's nova-api, nova-conductor
+and nova-scheduler services can run in kubernetes pods independently from the
+hypervisors they manage. In the other hand the nova-compute service is designed
+to be collocated with the hyperviors it manages. In the current design the EDPM
+Node serves as a hypervisor for Nova, therefore the nova-compute service needs
+to be deployed on EDPM nodes.
+
+Similar examples from other OpenStack compontents are ovn-metadata-agent, neutron-nicswitch agent, and in the HCI case the cinder-volume service.
```
stuggi, line 13: I think cinder-volume is in edge/DCN scenario on EDPM nodes, or has this changed? https://github.com/openstack/tripleo-heat-templates/blob/stable/train/roles/DistributedCompute.yaml#L28

spec/edpm-openstack-service-architecture.md
```diff
@@ -0,0 +1,323 @@
...
+
+The configuration options can be divided to two major parts:
+* configuration that is generated by the deployment engine. E.g. database
+  hostnames and message bus transport URLs
+* configuration that is provided by the human deployer. E.g. CPU allocation
+  ratio, PCI device spec and alias
+
+The existing podified control plane API, the OpenStackControlPlane CRD and the
+service sub CRDs, provides a simple way for the human deployer to pass
+configuration parameters to the OpenStack services. Each CRD that represents
```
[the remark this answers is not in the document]
gibizer: How I see it. We created k8s CRDs to model OpenStack services. Some of those services running as k8s Pods in the background some of those services running as podman containers on baremetal. The common denominator for me is that both representing OpenStack services not the technology implementing the execution of those services.

For some OpenStack services we will support NodeSelector (e.g. NovaAPI) and therefore those CRDs will have a corresponding field.  For others we wont (e.g. NovaExternalCompute) and therefore there won't be NodeSelector on those CRDs. But still both set of OpenStack services has the same oslo.config configuration interface so both services could be configured via the same oslo.config snippet via the CustomServiceConfig CRD parameter. For me this is a high level and consistent user interface as the use don't have to know which service is executed with witch technology when he or she wants to pass some configuration to it.

> We have plans to expose pass-through k8s options like podTemplate spec

Could you point me to the description of these plans?
The NodeSelector from the podTemplate does not make sense OpenStack services running on the EDPM node, so we don't expose (cannot expose) such configuration on those CRDs. Similarly if another podTemplate parameter does not make sense for OpenStack service on EDPM node then we won't expose them. But for others podTemplate parameters  (e.g. like restartPolicy) that make sense on EDPM node too, we can expose them and implement them.

spec/edpm-openstack-service-architecture.md
gibizer: Sorry that is my mixup. Fixing it...

spec/edpm-openstack-service-architecture.md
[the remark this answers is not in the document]
SeanMooney: i also dont thnkthat there is agreement aht we should expose

podTemplate, or other k8s constructs in the the CRDs.

we are developing an installer that uses Ansible and k8s internally to deploy OpenStack.
That does not mean we should expose all the options of either technology in our public API.

the pod template in partalare i feel would be incorrect to expose as that somewhat defeats the reason why we are developing operators in the first place.
we are trying to provide a simplified abstraction of the detail of the deployed service
exposing a podTemplate while it would simplify some aspect complicates others especially our support and mantaince of the deploy OpenStack.

spec/edpm-openstack-service-architecture.md
```diff
@@ -0,0 +1,326 @@
...
+phase 1 from the existing DataPlane model. This means that the DataPlane CRDs
+will not represent and therefore won't directly deploy any OpenStack services
+on the EDPM node. These CRDs still expose the status of the phase -1 and
+phase 0 deployment, so other operators can decide when to start deploying
+OpenStack service to the EDPM node.
+
+Define CRDs representing each OpenStack service type that needs to be run on
+the EDPM node (e.g. NovaExternalCompute, OVNMetadataAgent, etc). These CRDs
+are defined in and therefore reconciled by the respective service operator
+(e.g. nova-operator, ovn-operator). These OpenStack service CRDs will be
```
SeanMooney, line 125: nit: the OVNMetadataAgent woudl actully be reconsiled by the neutron-operator since its not actually part of ovn its part of neutron, just like the ovn ml2 driver is part of neutron. Its the neutron agent that implement the metadata proxy for the ovn backend.

spec/edpm-openstack-service-architecture.md
```diff
@@ -0,0 +1,326 @@
...
+created automatically based on the existence of DataPlaneRole and Node CRs and
+the user specified annotations on these CRs. The service operator
+will ensure that both the podified control plane dependencies
+(e.g. message bus) are deployed and the EDPM node phase 0 is finished before
+generates the service configuration and creates AnsibleEE CRs to deploy
+OpenStack service on the EDPM node.
+
```
SeanMooney, line 132: +1

spec/edpm-openstack-service-architecture.md
```diff
@@ -0,0 +1,326 @@
...
+spec:
+    dataPlane: openstack-edpm
+    nodeTemplate:
+       ansiblePort: 22
+...
+```
+The `novaTemplate` is removed from the `nodeTemplate` and a new user specified
+annotation `nova.openstack.org/cell` is added so the human deployer can
+define that this Role describe a set of EDPM nodes that are compute nodes
+connected to cell1.
```
SeanMooney, line 196: ack so ``nova.openstack.org/cell: nova-cell1`` is the minium
information we would require to be able to generate a default novaExternalCompute CR
since it contain a reference to the nova cell by name which also uniquely identifies the Nova instance as the cell name contains the nova instance name as a prefix.

this would result in the nova/libvirt service being deploy with our default rendered config and attaching the node to cell1.

spec/edpm-openstack-service-architecture.md
```diff
@@ -0,0 +1,326 @@
...
+    hostName: edpm-compute-1
+    node:
+        ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+    role: edpm-compute
+...
+```
+This DataPlaneNode has an `nova.openstack.org/cell` annotation that overrides
+the cell annotation of its role. So this EDPM node is a compute node connected
+to cell2.
+
```
SeanMooney, line 235: you also may prefer to have a role span multiple cells in which case you would just not set the annotation on the role and use it on the node but this make sense.

i would probably advise that OpenStackDataPlaneRole are cell alinged but there is no hard requirement either way so its good to all the felxablity to the deployer to choose how best to model there datacenter.

spec/edpm-openstack-service-architecture.md
```diff
@@ -0,0 +1,326 @@
...
+`CustomServiceConfig` is populated by the nova-operator. The
+`CustomServiceConfig` field is never set by the nova-operator it can be used
+by the human deployer.
+
+The following fields will be copied over from the DataplaneNode (or Role due to
+inheritance) and changes there will be propagated to the NovaExternalCompute CR
+* InventoryConfigMapName
+* NetworkAttachments
+* SshKeySecretName
+
```
SeanMooney, line 261: its not that it copied persay but its also probably good to note that the 
openstackDataplaneNode name will be used as the name of the NovaExternalCompute

in this case edpm-compute-0

we could technically not have them match and do something like
<nova cell>-<dataplane node> i.e. nova-cell1-edpm-compute-0
but that would d add complexity for little value.

un like the other filed that are copied however the CR name will be immutable.

spec/edpm-openstack-service-architecture.md
```diff
@@ -0,0 +1,326 @@
...
+The NovaInstance field is removed and the CellName field is redefined to hold
+the name of the NovaCell CR. The name of the NovaCell CR is a combination of
+the name of the cell from the NovaCellTemplate in the Nova CR and the name of
+the Nova CR. So it uniquely identifies a cell even if there are multiple Nova
+CRs exist in the namespace.
+
+The upstream nova project does not support moving computes between cells.
+Therefore the NovaCell field can only be set at creation and cannot be updated
+later. The dataplane-operator needs to implement a validation hook to prevent
+the related annotation to change after being set.
```
SeanMooney, line 276: hum so it could but i dont think it must do this.

instead we could have each service operator install a validating webhook for the dataplanenode/role CR that would just validate there annotation nothing else.

that way the dataplonenode does not need to have any knowledge of what the annotation are or do.

spec/edpm-openstack-service-architecture.md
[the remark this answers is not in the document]
stuggi: We have talked about that topic in the past in slack, and something we planned to look at in the next Q. 

I think we need to consider:
- In case of environment specifics or a bug, how can a user customize a deployment without the need to provide a new operator?
- how can support change settings to debug issues?
- have a generic way to customize e.g. timeouts [1]. for routes you use annotations [2] for this.

A generic way e.g. the rabbit operator, we consume, provides is to be able to override annotation and parts of the deployment spec of the service and StatefulSet. That's why we can consume it, can override the service spec to use a metallb loadbalancer, override the deployment spec with our needs. (**Note** it must not be the  full spec as it is a copy where everything is marked as optional). Using this pattern has the advantage to not have to add new parameters for every subset we realize at some point a user need to be able to customize. And we can still limit which parts. Like start with minimum required, like annotations. even our top level replicas could be then an override from the default.

[1] https://issues.redhat.com/browse/OSP-21715
[2] https://docs.openshift.com/container-platform/4.13/networking/routes/route-configuration.html#nw-configuring-route-timeouts_route-configuration
[3] https://github.com/rabbitmq/cluster-operator/blob/main/api/v1beta1/rabbitmqcluster_types.go#L164-L191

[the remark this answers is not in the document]
gibizer: Can we create a separate discussion for "How to expose podTemplate specific configuration for OpenStack service CRDs running in k8s pods?" I think my proposal quoted below shows that in the discussion of this PR we only need to care about that we *can* expose podTemplate specific parameters on those CRDs if we want. While not exposing them on service CRDs that doesn't or cannot support such parameters.:

> The NodeSelector from the podTemplate does not make sense OpenStack services running on the EDPM node, so we don't expose (cannot expose) such configuration on those CRDs. Similarly if another podTemplate parameter does not make sense for OpenStack service on EDPM node then we won't expose them. But for others podTemplate parameters (e.g. like restartPolicy) that make sense on EDPM node too, we can expose them and implement them.

I do have opinion about exposing the podTemplate directly on OpenStack service CRDs (so please ping me with that discussion if you start a separate one) but I think one way to help the EDPM architectural decision to move forward is to try to split out smaller independent decisions where possible.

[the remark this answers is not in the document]
gibizer: I've added a paragraph to emphasize that this spec does not try to prescribe that OpenStack service CRDs needs to be aligned regarding every field, e.g. NodeSelector (or podTemplate) might not be exposed for each service CRD.

gibizer (commented):
thank you for the feedback so far (keep them coming). I've updated the spec based on the comments.

spec/edpm-openstack-service-architecture.md
```diff
@@ -0,0 +1,348 @@
...
+    ansibleHost: 192.168.122.101
+    hostName: edpm-compute-1
+    node:
+        ansibleSSHPrivateKeySecret: dataplane-ansible-ssh-private-key-secret
+    role: edpm-compute
+...
+```
+This DataPlaneNode has an `nova.openstack.org/cell` annotation that overrides
+the cell annotation of its role. So this EDPM node is a compute node connected
+to cell2.
```
kk7ds, line 251: Ah, thanks I deleted another comment I had asking about this. So you can specify the cell per-role as a shortcut in the "I just have one cell" case, or set it per node. If I have different-enough roles for my two cells (i.e. high-spec and low-spec cells) then I could just tag those roles and not have to do it on my computes.

spec/edpm-openstack-service-architecture.md
```diff
@@ -0,0 +1,348 @@
...
+```yaml
+apiVersion: nova.openstack.org/v1beta1
+kind: NovaExternalCompute
+metadata:
+    name: edpm-compute-0
+spec:
+    dataplaneNodeName: edpm-compute-0
+    cellName: nova-cell1
+    customServiceConfig: |
+    # human deployer can set this to pass through service specific config
```
kk7ds, line 263: When and where? Patch this CR after it is created? (Yes, I see below).

Does the creation of this by the nova-operator immediately start the ansible jobs? (looks like it, once the node CR is ready per below). If so, I have to patch this after creation, which may re-trigger ansible job runs to update the values I set?

kk7ds (commented):
A couple comments inline, but this looks good to me from the nova perspective.

spec/edpm-openstack-service-architecture.md
SeanMooney: so i was wondering about that too. we could have a separate `nova.openstack.org/deploy`
annotation which would be True or False and default to false if not set.

or if we also implement the profile concept it could have a deploy field

either approach would allow you to have the service operator (nova in this case) do everything except deploy so you can examine the configs.

as it stands right now deploy true is our default os if we don't change that then yes the nova operator would deploy nova once the dataplane node ready condition is true.

gibizer: Good points. I only considered the case when the human wants to set debug=true for nova-compute on a single EDPM node as day 2 operation. In that case re-running the ansible is needed anyhow to change the config and restart the service. If the compute node specific config is needed during the first deploy already then that extra config either
* a) needs to come from the profile (see #42) 
* b) we need to model a deploy flag. 
  There is a deploy flag in DataPlaneNode.Spec.DeployStrategy.Deploy so we can simply use that as the source of information to set NovaExternalCompute.Spec.Deploy . I think we don't really need two separately settable flag, one compute specific annotation and one EDPM generic Spec field. I noted this now in the spec.

SeanMooney: we deffintly need two stpes since we are not going to reconsile until the datapalne node is ready right? unless your suggesting if the dataplen node is set to deploy false we would still generate all the config.

that would work if and only if we do not need to read info from the output of the datapalne operator.

i could see use getting PCI info or ips form it in the future so I'm not sure we can really on the dataPlaneNode.Spec.DeployStrategy for the service operators.

as it stand today we need the two levels for adoption to work properly.

spec/edpm-openstack-service-architecture.md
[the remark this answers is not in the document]
dprince: I think it is more than just documenting it in the spec, it is much larger than that. My main concern would be that this is an attempt to represent "external" resources on an entirely separate platform within a CRD that was designed specifically for k8s. It does not follow. I think it goes so far that I would say if we really intend to do this we should re-invision what the OpenStackControlplane CR represents, potentially removing some services, and these top level parameters which were intended to represent and manage services on k8s.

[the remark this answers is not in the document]
gibizer: I don't see why you think we cannot represent an externally running service as a k8s CRD. I agree that the externally running service should not be repredented by a k8s Pod. But why cannot be represented by a CRD?

There is a doc about [when to add a CRD](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#should-i-add-a-custom-resource-to-my-kubernetes-cluster). I don't see which point there invalidates our idea to represent an external service as a k8s CRD. Could you explain in more details what do you mean by "It does not follow" ?

In the other hand if we accept you point for a second that we cannot represent an external resource as a CRD then DataPlane, DataPlaneRole, DataPlaneNode and DataPlaneService are also invalid CRDs by your definition. What do you suggest there what should we do instead?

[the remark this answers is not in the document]
gibizer: If you are OK to represent external resources as CRDs in k8s, and your comment expresses that you are against including those external resource CRDs into the OpenStackControlPlane CRD, then I think there is a misunderstanding. The proposal in this document does not suggest that those external resource CRDs are part of the OpenStackControlPlane CRD. If you can point me which part of this doc causes the misunderstanding then I am happy to furter clarify it in the doc.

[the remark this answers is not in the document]
dprince: @gibizer It would help to explicitly call this out as not being part of the plan then. We've already had one service that implemented following the NovaExternalCompute model (Telemetry) show up on OpenStackControlplane. We have had previous conversations with Nova folks also state the intent was to treat it as part of the Controlplane. And I think others are confused as well.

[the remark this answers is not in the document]
dprince: @gibizer and to clarify your previous comment it is totally fine to represent external services with CRDs.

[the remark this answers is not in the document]
SeanMooney: The NovaExternalCompute is part of the contol plane but will not be part of the Nova cr and tasnsitvily will not be part of the OpenstackControlPlane CR.

The NovaCompute CRD will be part of the NovaCellTemplate which is part of the Nova CR and will be part OpenstackControlPlane CR.

The NovaCompute models a podifed nova-compute agent running on opensthif and will be used for the ironic and fake driver.  The NovaExternalCompute will as it is today be used for modelling nova-compute agents running on OpenstackDataPlaneNodes and be free standing form the OpenstackControlPlane CR. the proposal is to directly correlate it by name to the OpenstackDataPlaneNode i.e. and OpenstackDataPlaneNode called "compute-1" with the approatate annotation will result in the creation of a NovaExternalCompute named "compute-1" i would expect the telemetry operator to do the same for ceilometer-compute

create a ceilometerCompute Cr named compute-1 but that would not be in the OpenstackControlPlane CR

booxter:
I have nothing to add here except to confirm that this is in line with my thinking on next steps for OVNController and Neutron*Agents (and I believe this is also in line with Networking thinking.)

gibizer:
Moving this back to draft state as there is [a decision forming](https://issues.redhat.com/browse/OSP-25812?focusedId=22541705&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-22541705) that service operators should never run ansible via directly creating AnsibleEE CR. So the current proposal contradict with this decision https://github.com/openstack-k8s-operators/docs/pull/41/files#diff-b5fbfde815e7f3476b85b68cc2b58a93dce2cbbe2c94bc68a3e55cea541fdf5aR127-R131

gibizer:
This is not applicable any more based on the decision that sevice-operators are should not run or trigger any ansible execution, and instead DataPlaneService CRs will do that.

PR #42, opened 2023-06-13
[spec]Grouping nodes

gibizer:
Depends-on: #41


dprince:
@gibizer This came across my path today and I was a bit confused why it exists. We have at least 3 or 4 google docs discussing architectural options for Dataplane services. Is this meant to be a consolidation of the Google docs perhaps and you only want comments here now?

gibizer:
> @gibizer This came across my path today and I was a bit confused why it exists. We have at least 3 or 4 google docs discussing architectural options for Dataplane services. Is this meant to be a consolidation of the Google docs perhaps and you only want comments here now?

I'm trying to follow the pattern started in https://github.com/openstack-k8s-operators/docs/pull/31. Where after the initial discussion started to settle in the google docs the actual proposal is cleaned up and moved to this repo. This is intentionally in Draft state as I just finished writing this up as a cleanup of the google doc where the question / comments seems to be settled.

I'm traveling this week. So this and #41 will remain in Draft state at least until next week. If you have comments feel free to provide it here or in the google docs (whichever is more convenient for you) in the meantime. I will sync things up next week and we can also chat about the way forward.

spec/grouping-nodes.md
```diff
@@ -0,0 +1,146 @@
...
+DataPlaneRole or DataPlaneNode to select a single Profile per service type.
+
+### Example
+
+ComputeProfile with shared CPUs
+```yaml
+apiVersion: nova.openstack.org/v1beta1
+kind: ComputeProfile
+metadata:
+    name: shared-cpus
```
kk7ds, line 49: To be clear, it is likely this will not be as clear-cut for a real deployment, IIUC. Since you're embedding not just shared/dedicated cpus but *specific* ones, this would end up being something like `dell-r740-shared-cpus` or `hp-dl360g11-shared-cpus`, as different machines from different generations and configurations would end up with different config.

That doesn't change anything about what you're saying here, but I think it's important to note that we'd end up with a lot of ComputeProfile CRs. That's actually a *good* thing IMHO, because it helps to explain the scope of the explosion you mention above. It's not just that currently we'd need to have shared and dedicated versions of each neutron profile, but `hp-dl360g11-shared-dpdk` and `hp-dl360g11-dedicated-dpdk` and `hp-dl360g11-shared-nodpdk` and `hp-dl360g11-dedicated-nodpdk` ... *and* dell variants of each as well :)

spec/grouping-nodes.md
```diff
@@ -0,0 +1,146 @@
...
+
+OpenstackDataPlaneRole/edpm-compute
+```yaml
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneRole
+metadata:
+    name: edpm-compute
+    annotations:
+        nova.openstack.org/cell: nova-cell1
+        nova.openstack.org/compute-profile: shared-cpus
```
kk7ds, line 80: So since this is a dict and the key is (effectively) compute-profile, we're *not* talking about composing multiple `ComputeProfile`s. I think that's obvious, but maybe worth a note somewhere.

spec/grouping-nodes.md
```diff
@@ -0,0 +1,146 @@
...
+
+OpenstackDataPlaneNode/edpm-compute-1
+```yaml
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneNode
+metadata:
+    name: edpm-compute-1
+    annotations:
+        nova.openstack.org/compute-profile: pinned-cpus
+        neutron.openstack.org/network-profile: dpdk
```
kk7ds, line 108: So to be clear here, you've defined this as having role `edpm-compute` but have overridden each thing defined by the role, presumably just because you have to have *some* role right? It might be a better example to only override one of these two, or add a third profile that you still inherit from the role to make it clear that you can compose them that way.

That said, I hope people don't have to do this too much and that it is reserved for special one-off special-purpose machines.

kk7ds (commented):
This makes sense to be and addresses concerns we've had in previous discussions. A few supporting comments inline that may help with clarity in the long term.

spec/grouping-nodes.md
SeanMooney: right since we are not propsoing any composition of multiple profiles e.g. by making the annotation that resfences the profile a list
you would have to create a sperat profile for each hardwre/software combination


with that said we could evolve to enable composition if that was useful but I think there is merit in keeping this simple and just having an option 1:1 mapping rather then starting with many-to-one mapping and all of the precedence issues that would introduce.

spec/grouping-nodes.md
SeanMooney: I would not expect you would have to do this at the node level nessiarly but it would be good to have a multi-service example at the role level.

i.e. have OpenstackDataplaneRole edpm-nfv that uses the pinned-cpus nova profile and dpdk neutron profile.

that is how I would generally expect that to work.

with that said one use case for overriding at the node level is if the role had 
nova.openstack.org/compute-profile: <whatever>
telemetry.openstack.org/-profile: collector-rack1

but you wanted to use one of the nodes in that role to run extra services for that set of hosts

for example perhaps one of the nodes might also run collected/vector for monitoring aggreation

so that single node might have

nova.openstack.org/compute-profile: <whatever>
telemetry.openstack.org/-profile: collector-aggreator-rack1


that a bit contrived but the concept is simple all host in rack one would collect telemetry and forward it to the single node in rackone which is designated the aggrator that would then forward the telemetry for the entire rack to the central component.


unless you had something like that where all the node are basically the same expect one is slightly tweek i think it would make more sense to just have two dataplane Roles.

gibizer: I think your understanding is correct. The profile annotations in the node override the profile annotations inherited from the role.  Sean, you are probably right that it has a limited use case. Still because we have a Role - Node inheritance in the dataplane-operator we need to define what it means if the profile annotations are added to both the Node an the corresponding Role. So I choose the obvious meaning, the annotation on the more specific entity, the Node, overrides the annotation inherited from the more generic entity, the Role.

 I added the telemetry-profile to the example to show that we still inherit profiles from the Role that are note explicitly overridden on the Node.

spec/grouping-nodes.md
```diff
@@ -0,0 +1,155 @@
...
+
+Allow an EDPM node to belong to multiple groups by defining OpenStack service
+specific profiles. The DataPlaneRole could remain as is today but it would
+represent the grouping of the EDPM nodes only just from the generic
+hardware, host OS, infrastructure configuration perspective.
+
+Each service operator that has EDPM specific service CRD (e.g.
+NovaExternalCompute, OVNMetadataAgent) can define a Profile CRD (e.g.
+ComputeProfile, NetworkProfile) to describe the service specific configuration
+of a set of EDPM nodes. Then the human deployer can use annotations on the
```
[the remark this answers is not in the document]
SeanMooney: well the annotation and its type would technaily be up to the service operator to define

so if neutron wanted to support a list of profiles the neutron operator is free to do that.

we wanted to start with the simple case as for your node with both ovn and sriov you would create a thrid SRIOV_OVN profile that had the config for both.

granted that can lead to a large number of profiles if you need to multiple.

the current motivation to only having one profile si to avoid having to define any precedence relatantion or complex merging semantics. that does not mean the neutorn operator can support that we just didn't want to start from that state.

gibizer:
This is not applicable any more based on the decision that sevice-operators are should not run or trigger any ansible execution, and instead DataPlaneService CRs will do that. Such CRs forces to use DataPlaneRole for any kind of compute grouping. We might want to revisit the resulting Role explosion later.

PR #47, opened 2023-06-27
Document webhook logic reuse pattern

gibizer:


webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
 
 It is furthermore recommended to add support for running the operator locally with webhooks.  The pattern for doing so can be seen here: https://github.com/openstack-k8s-operators/cinder-operator/pull/155/files.
 
+### Reusing webhook logic
+
+Both the defaulting and the validation logic of the webhook are implemented in
+the service operators but to apply them meaningfully we need to call the that
+logic form the openstack-operator as the webhhooks will run there in a real
```
jpodivin, line 50: s/form/from/

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
 
 It is furthermore recommended to add support for running the operator locally with webhooks.  The pattern for doing so can be seen here: https://github.com/openstack-k8s-operators/cinder-operator/pull/155/files.
 
+### Reusing webhook logic
+
+Both the defaulting and the validation logic of the webhook are implemented in
+the service operators but to apply them meaningfully we need to call the that
```
jpodivin, line 49: That is enough.

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+1. OpenStackControlPlane.Default()
+2. FooSpec.Default()
+
+Therefore the service operator should not implement any defaulting logic in
+Foo.Default but all should be in FooSpec.Default() or in deeper calls
+
+**Validation**
+
+The validation webhook is more complex than the defaulting one as it:
+1. Needs to return errors to the caller than can be different depending on
```
jpodivin, line 93: s/than/that/

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+   where the webhook runs
+2. Needs to specific the path of the field in the validated structure that
```
jpodivin, line 95: s/specific/specify/

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+
+Still assume that Foo is a service CRD.
+
+In the service operator
+```golang
+func (r *Foo) ValidateCreate() error {
+	errors := r.Spec.ValidateCreate(field.NewPath("spec"))
+
+    if len(errors) != 0 {
+		novalog.Info("validation failed", "name", r.Name)
```
fmount, line 109: s/novalog/foolog ?

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+**Validation**
+
+The validation webhook is more complex than the defaulting one as it:
+1. Needs to return errors to the caller than can be different depending on
+   where the webhook runs
+2. Needs to specific the path of the field in the validated structure that
+   caused the validation error.
+3. Validation has multiple types, validation during create, update, or even
+   delete. The update validation has access to both the old and the new struct
+   value to be able to detect the change.
```
jpodivin, line 99: I would focus on this a bit more. Specifically, it would be a good idea to elaborate on requirements, that is MUST and MUST NOT, of each validation type.

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+                basePath.Child("myField"),
+                r.myField,
+                "should be shorter than 36 characters",
+            )
+        )
+    }
+    return errors
+}
+
+func (r *Nova) ValidateUpdate(old runtime.Object) error {
```
fmount, line 142: Same here (s/Nova/Foo)

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+2. FooSpec.ValidateCreate()
+3. FooSpec.ValidateMyFieldLength()
+
+When the webhook runs in the openstack-operator for an update the call chain
+is:
+1. OpenStackControlPlane.ValidateUpdate()
+2. FooSpec.ValidateUpdate()
+3. FooSpec.ValidateMyFieldLength()
+
+This validation chaining patter can be further extended inside the service
```
fmount, line 235: pattern

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+func (r *Foo) ValidateCreate() error {
+	errors := r.Spec.ValidateCreate(field.NewPath("spec"))
+
+    if len(errors) != 0 {
+		novalog.Info("validation failed", "name", r.Name)
+		return apierrors.NewInvalid(
+			schema.GroupKind{Group: "foo.openstack.org", Kind: "Foo"},
+			r.Name, errors)
+	}
+	return nil
```
jpodivin, line 114: Indentation in this example could be tidied up.

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+}
+
+func (r *FooSpec) ValidateCreate(basePath *field.Path) field.ErrorList {
+    // Implement the create validation here and return a list of errors.
+    var errors field.ErrorList
+    errors = append(errors,  r.ValidateMyFieldLength(basePath)...)
+
+    return error
```
jpodivin, line 122: s/error/errors/

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+func (r *Nova) ValidateUpdate(old runtime.Object) error {
+	errors := r.Spec.ValidateUpdate(old, field.NewPath("spec"))
+
+    if len(errors) != 0 {
+		novalog.Info("validation failed", "name", r.Name)
+		return apierrors.NewInvalid(
+			schema.GroupKind{Group: "foo.openstack.org", Kind: "Foo"},
+			r.Name, errors)
+	}
+	return nil
```
jpodivin, line 151: indentation

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+}
+
+func (r *NovaSpec) ValidateUpdate(old runtime.Object, basePath *field.Path) field.ErrorList {
+    // Implement the create validation here and return a list of errors.
+    var errors field.ErrorList
+    // You can also reuse specific validation from create validation here if
+    // needed
+    errors = append(errors,  r.ValidateMyFieldLength(basePath)...)
+
+	return errors
```
jpodivin, line 161: indentation

jpodivin (changes requested):
Overall I think this is a good pattern. But there are some stylistic issues that should be resolved.

jpodivin:
Also, linking to an existing implementation of the pattern could be useful.

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+
+When the webhook runs in the openstack-operator for an update the call chain
+is:
+1. OpenStackControlPlane.ValidateUpdate()
+2. FooSpec.ValidateUpdate()
+3. FooSpec.ValidateMyFieldLength()
+
+This validation chaining patter can be further extended inside the service
+operator if the service operator not just implement a top level service CRD but
+also sub CRDs.
```
fmount, line 237: I agree that we should find a common pattern for the upper layer interfaces, and when we reach the `Validate` function at `FooSpec` level then it's the <service>-operator itself that runs a set of validations based on the managed resources.
For instance I see the `ValidateMyFieldLength` at third level in the call chain above, but for me step 3 represents the Spec layer where `ValidateMyFieldLength` is just one of the set of validations I could run (and it depends on the service CRD and the exposed fields).
@gibizer Is my understanding correct?
Should we clarify that `ValidateMyFieldLength` is an example and from an high level point of view we can have a set of validations at Spec level and the generated resources?

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+        )
+    }
+    return errors
+}
+
+func (r *Nova) ValidateUpdate(old runtime.Object) error {
+	errors := r.Spec.ValidateUpdate(old, field.NewPath("spec"))
+
+    if len(errors) != 0 {
+		novalog.Info("validation failed", "name", r.Name)
```
fmount, line 146: ditto

webhooks.md
```diff
@@ -43,6 +43,199 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+		return apierrors.NewInvalid(
+			schema.GroupKind{Group: "foo.openstack.org", Kind: "Foo"},
+			r.Name, errors)
+	}
+	return nil
+}
+
+func (r *NovaSpec) ValidateUpdate(old runtime.Object, basePath *field.Path) field.ErrorList {
```
fmount, line 154: s/NovaSpec/FooSpec

webhooks.md
gibizer: Your understanding is correct. ValidateMyFieldLength is an example to show how to reuse a Validation that needs to be done both in case of Create and Update. There could be validations that only happen during Update (i.e. immutability check).

The pattern stops at the FooSpec level. The rest is just an example. I clarified this in the doc now.

webhooks.md
gibizer: I'm not sure what kind of requirements you are after. Do you have an example?

gibizer (commented):
Thank you for the feedback I was able to fix most of it

abays (approved):
/lgtm

webhooks.md
```diff
@@ -43,6 +43,201 @@ There are also additions/changes made to the `config` directory and to `main.go`
 
 It is furthermore recommended to add support for running the operator locally with webhooks.  The pattern for doing so can be seen here: https://github.com/openstack-k8s-operators/cinder-operator/pull/155/files.
 
+### Reusing webhook logic
+
+Both the defaulting and the validation logic of the webhook are implemented in
+the service operators but to apply them meaningfully we need to call the that
+from the openstack-operator as the webhhooks will run there in a real
```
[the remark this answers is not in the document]
gibizer: Good point. Done.

webhooks.md
```diff
@@ -43,6 +43,201 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+                basePath.Child("myField"),
+                r.myField,
+                "should be shorter than 36 characters",
+            )
+        )
+    }
+    return errors
+}
+
+func (r *Foo) ValidateUpdate(old runtime.Object) error {
```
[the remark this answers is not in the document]
gibizer: Good point I will adapt the pattern to this with one modification. 
// later
There is a complication so I'm glad you commented. When calling the service level FooSpec.ValidateUpdate(old, path) from openstack-operator we have to make sure the `old` is FooSpec not `OpenstackControlPlane`  so not just the basepath needs modification while calling down but the `old` is also needs to be scoped down. This means two casting is needed one on the openstack-operator code path and one in the service operator specific codepath.

Pattern is adapted now.

gibizer:
> Also, linking to an existing implementation of the pattern could be useful.

Done.

webhooks.md
```diff
@@ -43,6 +43,226 @@ There are also additions/changes made to the `config` directory and to `main.go`
...
+    oldFoo, ok := old.(*Foo)
+	if !ok || oldFoo == nil {
+		return apierrors.NewInternalError(fmt.Errorf("unable to convert existing object"))
+	}
```
stuggi, line 157: nit, indentation

stuggi (commented):
+1 from me

webhooks.md
jpodivin: Sure. I was thinking along the lines of:

> Create Validation MUST return error of the NewInvalid type, if it encounters any invalid section of the created data structure.

hjensas (commented):
LGTM.

webhooks.md
gibizer: Clarified it in the doc

gibizer:
The openstack-, ironic-, and nova-operator now aligned to this pattern. See https://github.com/openstack-k8s-operators/openstack-operator/pull/383 for details. So I think we can merge this doc now

jpodivin (approved):

fmount (approved):

PR #50, opened 2023-07-05
Update instructions for running local operator

ASBishop:
Update the instructions for creating a backup of the operator's CSV. The command needs to remove the /metadata/uid because that field is immutable.

### DIFF
--- a/running_local_operator.md
+++ b/running_local_operator.md
@@ -21,7 +21,7 @@ local operator will resolve its CR.
 
 ```
 oc get csv -n openstack-operators <your operator CSV> -o json | \
-  jq -r 'del(.metadata.generation, .metadata.resourceVersion)'  > operator_csv.json
+  jq -r 'del(.metadata.generation, .metadata.resourceVersion, .metadata.uid)'  > operator_csv.json
 ```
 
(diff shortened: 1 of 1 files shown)


running_local_operator.md
```diff
@@ -21,7 +21,7 @@ local operator will resolve its CR.
 
 ```
 oc get csv -n openstack-operators <your operator CSV> -o json | \
-  jq -r 'del(.metadata.generation, .metadata.resourceVersion)'  > operator_csv.json
+  jq -r 'del(.metadata.generation, .metadata.resourceVersion, .metadata.uid)'  > operator_csv.json
```
abays, line 24: But if it's the same instance of a CSV that you're backing up and restoring, then the UID should be the same.  If there were to be a conflict later, it would be because you had installed a new instance of the CSV since you originally made the backup, in which case I would think the backup is no longer relevant/needed.  Regardless, it doesn't hurt to remove the `.metadata.uid`, so this is fine.

abays (approved):
/lgtm

running_local_operator.md
abays: It's all good!  I'll merge it now.

PR #51, opened 2023-07-07
Configure IPs on isolated networks interfaces

stuggi:
IPs need to be configured on isolated network interfaces to have proper routing for the traffic coming via metallb vip into the cluster. If non is set, the return path back from the pod to the client will use the default GW since metallb does not do NAT and there is no association to the isolated network interface. nmstate right now also does not allow to create a route like `ip route add 192.168.122.0/24 dev enp6s0`. It gets blocked on validation since IPv4 is disabled for the interface.


stuggi:
this was previously part of the doc and I wrongly removed it

abays (approved):
/lgtm

PR #52, opened 2023-07-21
[envtest] Fix helper links after reorg

karelyatin:
The test helpers were restructured with [1].
Update doc to reflect it.

[1] https://github.com/openstack-k8s-operators/lib-common/pull/301


gibizer (approved):
Thank you @karelyatin !

PR #54, opened 2023-08-08
Add documentation on using the ceph-hci-pre service

fultonj:
Document how to use the ceph-hci-pre service when deploying HCI to configure the firewall for Ceph services.

Jira: OSP-25827
Depends-On: https://github.com/openstack-k8s-operators/dataplane-operator/pull/341

### DIFF
--- a/hci.md
+++ b/hci.md
@@ -67,8 +67,9 @@ next sections will be required.
 
 ### Shorten the Service list
 
-Shorten the services list so that it only has the services called
-`configure-network` and `validate-network`.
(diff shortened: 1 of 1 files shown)


fultonj:
We don't want to merge this until it's depends-on merges. Otherwise it would describe a service which is not yet available.

https://github.com/openstack-k8s-operators/dataplane-operator/pull/341

fao89 (commented):
+2

rebtoor (approved):

fmount (approved):
LGTM

PR #55, opened 2023-08-10
Document EDPM ceph-client service

fultonj:
It is now necessary to include the ceph-client service when configuring Ceph or HCI so update documents accordingly.

Jira: OSP-25826

### DIFF
--- a/hci.md
+++ b/hci.md
@@ -101,7 +101,7 @@ firewall ports necessary to run the Ceph RBD block storage service
 from the EDPM storage network (`172.18.0.0/24`) because of the
 following defaults for the `edpm_ceph_hci_pre_storage_ranges` and
 `edpm_ceph_hci_pre_enabled_services` Ansible variables.
-```
+```yaml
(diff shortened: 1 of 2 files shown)


fmount (approved):
LGTM

fmount:
We can land this patch right after https://github.com/openstack-k8s-operators/dataplane-operator/pull/350

hci.md
```diff
@@ -496,9 +496,17 @@ spec:
...
+        - ceph-client
         - configure-os
         - run-os
 ```
+In addition to restoring the default service list, the `ceph-client`
+service is added after `install-os` and before `configure-os`. The
+`ceph-client` service configures EDPM nodes as clients of a Ceph
+server by distributing the files, which Ceph clients use, which
+were made available as described in the "Add ExtraMounts" section
+of the document.
```
fao89, line 508: 👍🏽

fao89 (approved):

PR #56, opened 2023-08-17
Update Ceph/HCI documents for dataplane configMaps update

fultonj:
It is no longer supported [1] to use a `nova` dictionary within a `OpenStackDataPlane` CR to pass configuration overrides to Nova. Instead the overrides should be stored in a `ConfigMap` and a custom version of the `nova service` should reference it using `configMaps` within an `OpenStackDataPlaneService` CR.

Re-write Ceph/HCI documents so that reader can get the same configuration as before but using the new method.

[1] https://github.com/openstack-k8s-operators/dataplane-operator/pull/337

### DIFF
--- a/ceph.md
+++ b/ceph.md
@@ -219,40 +219,71 @@ to be considered secret.
 
 ## Configure Nova
 
-Use the `OpenStackDataPlane`
-[NovaTemplate](https://openstack-k8s-operators.github.io/dataplane-operator/openstack_dataplanerole/#novatemplate)
-to pass a `customServiceConfig`.
-
+Create a ConfigMap with content to be added to /etc/nova/nova.conf.d/,
+inside the nova_compute container, so that Nova uses Ceph RBD.
 ```yaml
-apiVersion: dataplane.openstack.org/v1beta1
-kind: OpenStackDataPlane
-spec:
-<snip>
-  roles:
-    edpm-compute:
-    <snip>
-        nova:
-          cellName: cell1
-          customServiceConfig: |
-            [libvirt]
-            images_type=rbd
-            images_rbd_pool=vms
-            images_rbd_ceph_conf=/etc/ceph/ceph.conf
-            images_rbd_glance_store_name=default_backend
-            images_rbd_glance_copy_poll_interval=15
-            images_rbd_glance_copy_timeout=600
-            rbd_user=openstack
-            rbd_secret_uuid=$FSID
-          deploy: true
-          novaInstance: nova
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ceph-nova
+data:
+  03-ceph-nova.conf: |
+    [libvirt]
+    images_type=rbd
+    images_rbd_pool=vms
+    images_rbd_ceph_conf=/etc/ceph/ceph.conf
+    images_rbd_glance_store_name=default_backend
+    images_rbd_glance_copy_poll_interval=15
+    images_rbd_glance_copy_timeout=600
+    rbd_user=openstack
+    rbd_secret_uuid=$FSID
 ```
+The filename, e.g. 03-ceph-nova.conf, must match "*nova*.conf" and
+the files are evaluated by Nova alphabetically (e.g. 01-foo-nova.conf
+is processed before 02-bar-nova.conf).
+
 The `$FSID` value above should contain the actual FSID as described
 in the "Create a Ceph Secret" section. The FSID itself does not need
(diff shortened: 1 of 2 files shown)


fmount (approved):
Thanks John for the update.

ceph.md
```diff
@@ -219,40 +219,71 @@ to be considered secret.
...
+    images_type=rbd
+    images_rbd_pool=vms
+    images_rbd_ceph_conf=/etc/ceph/ceph.conf
+    images_rbd_glance_store_name=default_backend
+    images_rbd_glance_copy_poll_interval=15
+    images_rbd_glance_copy_timeout=600
+    rbd_user=openstack
+    rbd_secret_uuid=$FSID
 ```
+The filename, e.g. 03-ceph-nova.conf, must match "*nova*.conf" and
```
fmount, line 242: This is interesting, I thought that an order would have been adopted within the same `Secret/ConfigMap` (e.g. I have `01-foo-nova.conf` and `03-ceph-nova.conf` within the same `ConfigMap` and I want to make sure an order is respected when processing the content of these two keys), but I didn't know this order is actually respected among multiple `ConfigMaps`.
For instance, even though I have `configmap-01` providing `03-ceph-nova.conf`, and `configmap-02` providing `01-nova-conf.conf`, I assume nova is able to process `01-nova-conf.conf` before `03-ceph-nova.conf`.
I understand that the config can be provide in a mixed form of ConfigMaps and Secrets, so it makes sense moving the logic in the nova controller.
This doesn't change anything in this doc, hence not something that should be explained here.

ceph.md
```diff
@@ -219,40 +219,71 @@ to be considered secret.
...
+```yaml
+---
+apiVersion: dataplane.openstack.org/v1beta1
+kind: OpenStackDataPlaneService
+metadata:
+  name: nova-custom-ceph
+spec:
+  label: dataplane-deployment-nova-custom-ceph
+  configMaps:
+    - ceph-nova
```
fmount, line 263: +1 the configmap is referenced here

hci.md
```diff
@@ -558,6 +521,18 @@ for the hypervisor. In an IOPS-optimized cluster performance can be
...
+Use `oc edit OpenStackDataPlaneService/nova-custom-ceph` to add
+`resereved-memory-nova` to the `configMaps` list.
+
+```yaml
+---
+kind: OpenStackDataPlaneService
+<...>
+spec:
+  configMaps:
+  - ceph-nova
```
fmount, line 533: @fultonj not sure I get why we can't put the two contents (`03-ceph-nova.conf` and `04-reserved-memory-nova.conf`) within a single `ConfigMap` instead of having both ceph-nova` and `reserved-memory-nova`

hci.md
```diff
@@ -494,62 +476,43 @@ spec:
...
-            rbd_secret_uuid=$FSID
-          deploy: true
-          novaInstance: nova
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: resereved-memory-nova
+data:
+  04-resereved-memory-nova.conf: |
```
fmount, line 511: s/resereved/reserved

fultonj: fixed

hci.md
fultonj: We could combine them and either results in the same configuration. I separated them because:

- 03-ceph-nova.conf is from the ceph document and needed for external Ceph only
- 04-reserved-memory-nova.conf is from the HCI document and needed only for the 

Combining them  would not let me reference the first and have to document twice. Either one produces a valid config. One could argue it helps organize logically separate configurations.

ceph.md
fultonj: > This is interesting, I thought that an order would have been adopted within the same `Secret/ConfigMap` (e.g. I have `01-foo-nova.conf` and `03-ceph-nova.conf` within the same `ConfigMap` and I want to make sure an order is respected when processing the content of these two keys), but I didn't know this order is actually respected among multiple `ConfigMaps`. For instance, even though I have `configmap-01` providing `03-ceph-nova.conf`, and `configmap-02` providing `01-nova-conf.conf`, I assume nova is able to process `01-nova-conf.conf` before `03-ceph-nova.conf`. I understand that the config can be provide in a mixed form of ConfigMaps and Secrets, so it makes sense moving the logic in the nova controller. This doesn't change anything in this doc, hence not something that should be explained here.

I was focusing about how Nova evaluates them once they land on the host. I agree we don't want to dive into the implementation details on the deployment side here.

PR #57, opened 2023-08-28
Update Ceph documentation for recent changes

fultonj:
Document using multiple Ceph keyrings in a secret CR. Document where ceph files are stored on EDPM nodes. Update Manila example to include missing pools.
Update ceph-nova custom service example with playbook as it has recently changed.

Jira: OSP-27777

### DIFF
--- a/ceph.md
+++ b/ceph.md
@@ -51,10 +51,13 @@ Create a cephx key which OpenStack can use to access the pools.
 ```shell
 cephadm shell -- \
    ceph auth add client.openstack \ 
-     mgr 'allow *' \
+     mgr 'allow rw' \
 	 mon 'allow r' \
 	 osd 'allow class-read object_prefix rbd_children, allow rwx pool=vms, allow rwx pool=volumes, allow rwx pool=images'
 ```
+If Manila is enabled in the OpenStack controlplane, then add `allow
+rwx pool=cephfs.cephfs.meta, allow rwx pool=cephfs.cephfs.data` to the
+end of the command above.
 
 Export the cephx key and Ceph configuration file.
 ```shell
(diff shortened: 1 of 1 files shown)


ceph.md
```diff
@@ -55,6 +55,9 @@ cephadm shell -- \
 	 mon 'allow r' \
 	 osd 'allow class-read object_prefix rbd_children, allow rwx pool=vms, allow rwx pool=volumes, allow rwx pool=images'
 ```
+If Manila is enabled in the OpenStack controlplane, then add `allow
```
fmount, line 58: +1 given we already included the `gs volume create` on L47

fmount: Also, I'm wondering if in L54 would be enough having `mgr 'allow rw'` instead of `*`. In my env (which is the same key used in NG CI) we have [1] and we could think about updating L54 accordingly.
However, I see in TripleO we have [2], so I'm not sure which service requires `allow *` instead of `rw`.
@fultonj wdyt?

[1] https://github.com/openstack-k8s-operators/install_yamls/blob/main/scripts/gen-ceph-kustomize.sh#L177 
[2] https://github.com/openstack/tripleo-ansible/blob/master/tripleo_ansible/roles/tripleo_run_cephadm/tasks/build_keys.yml#L32

ceph.md
```diff
@@ -152,10 +184,14 @@ When a CR containing the above is created, an Ansible pod
...
 and copies them to the EDPM host using the
 [edpm_ceph_client_files](https://github.com/openstack-k8s-operators/edpm-ansible/tree/main/roles/edpm_ceph_client_files)
-Ansible role. The Nova containers then have a copy of /etc/ceph
+Ansible role. The Nova containers then have a copy of `/etc/ceph`
 that they can use to acccess the cephx key and Ceph configuration
 file.
 
+Though containers hosted on EDPM nodes use the path `/etc/ceph`
+to access the Ceph client files, they are stored in
+`/var/lib/openstack/config/ceph` on the EDPM nodes.
```
fmount, line 193: +1 on using a different path in the host, which is consistent with what we did in TripleO [1]

[1] https://github.com/openstack/tripleo-heat-templates/blob/master/deployment/cephadm/ceph-client.yaml#L58

ceph.md
fultonj: I'll update it to `rw` instead of `*`. thanks

PR #58, opened 2023-08-30
[envtest]Document debugging with delve

gibizer:


envtest.md
```diff
@@ -279,6 +279,12 @@ Here is an example launch configuration for vscode:
 Note that you at least needs to run the test with the make target to get the
 binaries downloaded to the local ./bin folder.
 
+### Debug with delve
+You can also use the command line debugger to debug the envtest execution:
+```shell
+KUBEBUILDER_ASSETS=$(pwd)/bin/k8s/1.26.1-linux-amd64 OPERATOR_TEMPLATES=$(pwd)/templates dlv test ./test/functional/...
```
Akrog, line 285: Looks like Nova is the only operator calling their testing directory `test` the rest (glance, cinder, keystone, openstack) call it `tests`.

gibizer: I guess `test` comes from https://github.com/golang-standards/project-layout but I don't have strong opinion.

Akrog: I do have strong opinions 😅
We should be consistent across all operator projects, and if there is a standard layout, like here, follow it (even if I don't like calling the directory in singular).
So I guess I'll be updating the cinder-operator directory to `test`.

bogdando (approved):

PR #59, opened 2023-09-01
Add external RGW backend for Object-Store service

fmount:
This patch introduces a section that describes how to configure a `RadosGW` instance as a backend for the 
`OpenStack` `Object-Store` service.

### DIFF
--- a/ceph.md
+++ b/ceph.md
@@ -381,6 +381,117 @@ spec:
                     cephfs_protocol_helper_type=CEPHFS
 ```
 
+## Configure Swift with a RGW backend
+
+It's possible to configure an external Ceph Object Gateway (RGW) as an Object
+Store service. Users can rely on the `Swift` client tool to interact with the
+object store service.
+To authenticate with the external RGW service, you must configure RGW to verify
+users and their roles in the Identity service (keystone).
+Run the following commands on an already deployed OpenStack control plane to
+create users and roles as they will be used by the RGW instances to interact with
+keystone.
+
+```bash
+openstack service create --name swift --description "OpenStack Object Storage" object-store
+openstack user create --project service --password $SWIFT_PASSWORD swift
+openstack role create swiftoperator
+openstack role create ResellerAdmin
+openstack role add --user swift --project service member
+openstack role add --user swift --project service admin
+
+export RGW_ENDPOINT=192.168.122.3
+for i in public internal; do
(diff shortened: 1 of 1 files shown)


ceph.md
```diff
@@ -381,6 +381,113 @@ spec:
...
+spec:
+  backend_service: rgw.rgw
+  frontend_port: 8080
+  monitor_port: 8999
+  virtual_ip: $STORAGE_NETWORK_VIP
+  virtual_interface_networks:
+  - $STORAGE_NETWORK
+```
+
+- Replace `$HOST1`, `$HOST2`, ..., `$HOSTN` with the number of `CephStorage`
```
fultonj, line 480: Let's not use `CephStorage` because that had special meaning in TripleO and we don't want anyone to think some form of it has lived on in the new deployment method.

"with the number of Ceph nodes where the ..."

fultonj (changes requested):
Thanks Francesco. Overall this is good. Just requesting a few small changes.

ceph.md
```diff
@@ -381,6 +381,113 @@ spec:
...
+```bash
+openstack service create --name swift --description "OpenStack Object Storage" object-store
+openstack user create --project service --password $SWIFT_PASSWORD swift
+openstack role create swiftoperator
+openstack role create ResellerAdmin
+openstack role add --user swift --project service member
+openstack role add --user swift --project service admin
+
+for i in public internal; do
+    openstack endpoint create --region regionOne object-store $i http://<RGW_ENDPOINT>:8080/swift/v1/AUTH_%\(tenant_id\)s;
```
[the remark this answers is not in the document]
fmount: Thanks John for the review here. Yeah, I'll do an example and suggest the strategy that should be adopted when the RGW_ENDPOINT is chosen.

ceph.md
fmount: Ack I'm going to `s/CephStorage/Ceph` here. Let's avoid confusion or any call to other projects.

fmount: Done

ceph.md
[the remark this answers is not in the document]
fmount: Ok I updated it [1]: let me know your thoughts about the last change.

[1] https://github.com/openstack-k8s-operators/docs/compare/0f1d5ba04f6960849a5e2eff737174873a6295b9..19a8d493bc3670eecc55fa949a4a6fb9cb923923

PR #61, opened 2023-09-20
Update Ceph docs for OpenStackDataPlaneNodeSet

fultonj:
The OpenStackDataPlaneNodeSet CRD has been introduced to the data plane operator so the Ceph documentation needs to be updated to work with the new CRD.

Also, ceph-hci-pre needs to come before configure-os, as the latter is what actually applies the firewall rules that the former generates.

### DIFF
--- a/ceph.md
+++ b/ceph.md
@@ -161,24 +161,23 @@ spec:
             mountPath: "/etc/ceph"
             readOnly: true
 ```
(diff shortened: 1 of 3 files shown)


abays (approved):
/lgtm

fmount (approved):
Thanks John, looks good!

PR #62, opened 2023-09-25
Update networking doc for service override change

stuggi:
Jira: OSP-26690

Depends-On: https://github.com/openstack-k8s-operators/openstack-operator/pull/457

### DIFF
--- a/networking.md
+++ b/networking.md
@@ -250,113 +250,75 @@ spec:
 
 ### Configure exposing services to isolated networks
 #### Service API component
-Each API component of a service supports exposing their endpoint to an isolated network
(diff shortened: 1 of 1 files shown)


networking.md
```diff
@@ -250,113 +250,75 @@ spec:
...
-            specified.
-          type: string
-      required:
-      - endpoint
-      - ipAddressPool
-```
+For services which get registered to different OpenStack endpoint types a hash `map[service.Endpoint]service.RoutedOverrideSpec` is used, where the key is either `public` or `internal`.
 
-Example:
+Example: To expose the internal keystone endpoint using a MetalLB k8s LoadBalancer via the service override the `LoadBalancer` type in the spec and required annotations in the metadata has to be set:
```
abays, line 257: I would put a comma after `LoadBalancer via the service override`.  It helps the reader parse the long sentence.

stuggi: done

abays (approved):
/lgtm

PR #63, opened 2023-09-25
Fix external RGW doc to point to keystone internal endpoint

fmount:
EDPM nodes resolve the internal endpoint associated with a given service. This change fixes the doc where we previously suggested to add the "ceph config global set rgw_keystone_url" pointing to the identity public endpoint.


fultonj (approved):
Thanks Francesco

abays (approved):
/lgtm

PR #69, opened 2023-10-12
Update HCI documentation to use pre/post Ceph deployments

fultonj:
Use servicesOverride feature of OpenStackDataPlaneDeployment to make HCI documentation more clear. Instead of having human operator edit the services list before creating a second deployment, provide two deployment examples which can be run before and after Ceph is deployed.

This patch links to samples which will be introduced by the Depends-On patch so please merge it first.

Jira: OSP-29040
Depends-On: https://github.com/openstack-k8s-operators/dataplane-operator/pull/464

### DIFF
--- a/hci.md
+++ b/hci.md
@@ -39,8 +39,9 @@ spec:
 Only the services which are on the list will be configured.
 
 Because we need to deploy Ceph on an EDPM node after the storage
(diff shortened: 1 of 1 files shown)


fmount:
+2 from me

PR #71, opened 2023-10-17
Document dataplane-operator scale down for local work

fultonj:
Add some missing information about how to determine the CSV of the dataplane-operator so that it can be scaled down for local development.


abays (approved):
/lgtm

fao89 (approved):

PR #73, opened 2023-10-20
Document adding commands to pod containers

ASBishop:


adding_shell_commands.md
```diff
@@ -0,0 +1,81 @@
+# Adding shell commands to pod containers
```
stuggi, line 1: another option is to use the `oc debug` command with the `--image` parameter to use a different container image. `oc debug` will start a new container in the context of the pod, e.g.  `oc debug cinder-api-769d687f7-rg5bq -n openstack --image quay.io/fedora/fedora` .  Depending of the user running the pod/container you could then use `dnf` to install packages. 
install iputils will not work as it would required the `SETFCAP` cap to be able to install the pkg. More details in  https://access.redhat.com/solutions/6607631 . In this case a prebuilt image with the required tools could be used. But in general this is also a way could be used if tools are missing.

ASBishop: Sure, I guess people could try that as well. For me, I want to do the debug work in the actual container, not alongside it in a separate debug container.

slagle (approved):

PR #77, opened 2023-11-28
Clarify that service_config is for oslo.config services

booxter:
OVN has no use for kolla config file management. Kolla just gets in the way by putting additional requirements as to permissions on top of what OVN services actually may need.


karelyatin (commented):
/lgtm

PR #78, opened 2023-12-05
Add note about creds for skopeo for OpenStack operator image build

abays:


image_build.md
```diff
@@ -21,6 +21,18 @@ make catalog-build catalog-push
 
 Also check the operator-sdk quickstart at https://sdk.operatorframework.io/docs/building-operators/golang/quickstart/
 
+### OpenStack operator note
+
+Building the OpenStack operator images locally requires `skopeo`, which itself must log into quay.io during its execution.  
+Thus this requires authentication.  By default, `skopeo` looks for login credentials in a location that may or may not currently
```
dprince, line 27: I think all of our quay.io images are public. The only authentication that should be required is for the few registry.redhat.io images we have (I think)?

dprince (commented):
Perhaps it is worth qualifying these steps as something folks might only need on older RHEL8 machines. I do not have to complete these steps on Fedora

image_build.md
abays: It does seem to matter for quay.io images as well in my environment:

```
FATA[0000] initializing source docker://quay.io/openstack-k8s-operators/barbican-operator-bundle:96ca1e05d8ad24687cc0e708588e2bfe50f3c1cb: getting username and password: 1 error occurred:
	* reading JSON file "/run/containers/1000/auth.json": open /run/containers/1000/auth.json: permission denied
```

abays:
> Perhaps it is worth qualifying these steps as something folks might only need on older RHEL8 machines. I do not have to complete these steps on Fedora

Yes, I will update the note to reflect this

PR #85, opened 2024-01-03
[network] remove storage_mgmt network from ceph

### LINKED ISSUE #81: Missing a VLAN for storage management
In https://github.com/openstack-k8s-operators/docs/blob/main/images/network_diagram.jpg there's a "storage mgmt" in the Ceph storage but nothing connects to it. Comparing it with the devpreview2 docs, It seems noth the OSP compute node and something in OCP worker nodes should connect to it as a VLAN.

stuggi:
Closes #81


fultonj:
I'm fine with this since the diagram is generic and no OpenStack services need to connect to the storage_mgmt network. Technically, in the HCI scenario, we need the IPAM in OCP to be aware of the storage_mgmt network so that EDPM nodes can be configured with it. However, that detail does not need to be captured in the diagram.

https://github.com/openstack-k8s-operators/docs/issues/81#issuecomment-1875374600

flozanorht:
Wouldn't it be better having a line from compute nodes to the storage network to show who uses it? After all, it makes no sense having something in a generic diagram that no one uses or needs. If it's there, something has to show the need for it.

fultonj:
> Wouldn't it be better having a line from compute nodes to the storage network to show who uses it? 

There is. Please have another look at the new diagram.

https://raw.githubusercontent.com/openstack-k8s-operators/docs/97b732cef271742a85daa1b712b4132097703d35/images/network_diagram.jpg

> After all, it makes no sense having something in a generic diagram that no one uses or needs. If it's there, something has to show the need for it.

This is a network someone would use. It's not HCI but I'd consider it to be the "external ceph" diagram.

flozanorht:
> There is. Please have another look at the new diagram.

Oh I didn't notice you made changes, I misread as dismissing my issue with no changes. Thanks a lot!

You mention an "external ceph" diagram.. I cannot find it.

fultonj:
> > There is. Please have another look at the new diagram.
> 
> Oh I didn't notice you made changes, I misread as dismissing my issue with no changes. Thanks a lot!

I didn't make the changes. They were in Martin's original PR.

> You mention an "external ceph" diagram.. I cannot find it.

There is one diagram right now. It covers a scenario where ceph is hosted externally, i.e. not on EDPM nodes.

flozanorht:
"You" I meant your team or all the OSP team. And thanks for confirming there's no other diagram I'd be missing.

Issue #84, opened 2024-01-02
How external traffic reaches VMs inside a compute node?


flozanorht:
The diagram https://github.com/openstack-k8s-operators/docs/blob/main/images/network_diagram.jpg gives no clue about it. As the only "external" network there is the OpenShift cluster network (or whenever OpenShift ingress controllers and default LB services get their VIPs) it seems it has to be forwarded by OpenShift worker nodes, but how?

stuggi:
@slawqo or @karelyatin can better answer that.

karelyatin:
I think i am seeing this diagram first time. But yes you are right it's missing external network references on compute node so that needs to be fixed.

With default dvr configuration, external interface is required on the compute nodes. Right now with the validated architecture dev setup we are using ctlplane network as external n/w for FIP traffic. So VMS with FIPs get's accessible over this network. And SNAT for VMS without FIP goes via ocp worker nodes over the same ctlplane interface(enp7s0).

stuggi:
@karelyatin so for a general nw architecture it would be like https://github.com/openstack-k8s-operators/docs/pull/86

PR #86, opened 2024-01-04
[network] add OSP external network for computes to diagram

stuggi:
Closes: https://github.com/openstack-k8s-operators/docs/issues/84


karelyatin:
Looks good for compute nodes but i think we need that also on OCP Worker nodes too and ovn-controller pods will utilize that for SNAT traffic

stuggi:
> Looks good for compute nodes but i think we need that also on OCP Worker nodes too and ovn-controller pods will utilize that for SNAT traffic

right, I'll update the PR for that

stuggi:
updated, @karelyatin please have another look

PR #87, opened 2024-01-08
Align glanceAPI to the new CRD and update the associated samples

fmount:
Now that glance has a new CRD and API definition, we need an updated doc to avoid confusion between the old and the new version.


abays (approved):
/lgtm

PR #88, opened 2024-01-12
[TLS] internal tls documentation

stuggi:
Jira: OSPRH-2190

### DIFF
--- a/tls.md
+++ b/tls.md
@@ -88,7 +88,7 @@ This bundle can be used to be direct mounted into a deployment to the environmen
 
 ### TLS internal communication
 
-Each service operator creates for its APIs one or multiple k8s services. If in the `OpenStackControlPlane` CR gets configured to have tls for internal endpoints enabled:
+Each service operator creates one or multiple k8s services for its APIs. If in the `OpenStackControlPlane` CR is configured to have tls for internal endpoints enabled:
 
 ```yaml
 apiVersion: core.openstack.org/v1beta1
--- a/tls.md
+++ b/tls.md
@@ -29,10 +29,10 @@ spec:
         enabled: false
       public:
         enabled: true
-    caSecretName: myAdditionalCACerts
+    caBundleSecretName: myAdditionalCACerts
 ```
 
-Using the `caSecretName` parameter a secret can be referenced containing any additional CA certificates, which should be added to the CA bundle.
+Using the `caBundleSecretName` parameter a secret can be referenced containing any additional CA certificates, which should be added to the CA bundle.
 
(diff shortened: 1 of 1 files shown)


fao89 (approved):

tls.md
```diff
@@ -88,4 +88,58 @@ This bundle can be used to be direct mounted into a deployment to the environmen
...
+        secretName: cert-public-svc
+    # secret holding the tls-ca-bundle.pem to be used as a deploymend env CA bundle
+    caBundleSecretName: combined-ca-bundle
+```
+
+Single, not endpoint specific service:
+```yaml
+  tls:
+    caBundleSecretName: combined-ca-bundle
+    secretName: cert-nova-novncproxy-cell1-public-svc
```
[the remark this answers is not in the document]
stuggi: this service does not have internal/public. its exposed using only a single k8s service

tls.md
```diff
@@ -88,4 +88,58 @@ This bundle can be used to be direct mounted into a deployment to the environmen
...
+```
+
+As mentioned above, the CA bundle will be mounted as the global CA bundle for the deployment container. The service secrets will be used for the service providing the api. Depending on how the service gets started (using kolla or not), the cert and key get either mounted to a directory and using kolla to put into its final place, or direct mounted to where the service expects it.
```
gibizer, line 135: I just noticed that we have both in nova-api. The CA boundle is mounted directly, while the server cert and private key is copied via kolla. Is this intentional?

gibizer: Thinking more about it what bothers me is the two ways of putting TLS "stuff" in place when kolla is involved as the CA bundle always mounted directly regardless of kolla. 
Anyhow it is not a big deal just caused a bit of confusion when looked at nova-api statefulset, saw the tls sercret mounted to /etc/pki and I (wrongly) assumed that mount put in place all the TLS material.

stuggi: yes, this is expected. the ca bundle is global readable, so its mounted direct. the service certs, especially the key, need restricted permissions. Right now we have mixed service configurations throughout the service operators. some pods run as root, some as the service user. if we'd mount them direct, depending on the current implementation of the service operator, it would required to set the FSGroup. Therefore we decided to keep using kolla for moving the service certs to the destination and set the correct owner, if kolla is currently used and review this later. we did not want to delay the tls topic to first get this aligned as this can be done as a follow up.  it is transparent and could even be done after GA.

gibizer: Thanks for the explanation.

Deydra71 (approved):

PR #89, opened 2024-01-16
Add version_updates.md (OpenStackVersion docs)

dprince:


version_updates.md
```diff
@@ -0,0 +1,120 @@
...
+If a situation arises where we need to update one service before
+another we can use OpenStackVersion to drive this.
+
+## Things currently not supported
+
+ * Rolling back to a prior openstack version directly via OpenStackVersion.
+   A rollback would require the administrator to install an old CSV,
+   and patch various CRs to update the containerImages among other things.
+ * Major upgrades are not targeted at this time. However the OpenStackVersion
+   CR may be useful as a means to help drive those as well.
```
gibizer, line 120: Does any CRD changes mean a major update?

version_updates.md
```diff
@@ -0,0 +1,120 @@
...
+## General Update Workflow
+
+OpenStack is installed by following the normal installation procedure. When
+an OpenStackControlplane resource is created in any given namespace an
+OpenStackVersion resource will be immediately created when the OpenStackControlPlane
+is reconciled (1 OpenStackVersion per namespace).
+
+When a new set of Operators is installed via OLM (either manually or automatically)
+the OpenStackVersion resources will reconcile and a new 'AvailableVersion' will
+be set on each CR.
```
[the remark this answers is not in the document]
dprince: Container images are never really deleted however if you install a new set of openstack operator's (the entire set) it is very likely that all new image defaults will exist. The old image references will persist on any existing/deployed OpenStackControlplane and will not change or be deleted in anyway until a user kicks off the minor update process. In summary, installing new operators via OLM gives you new image defaults, and the option to perform a minor update as well.
As for manipulations only the image versions are getting patched via OpenStackVersion. Once we make a meaningful CRD change that requires a new version of OpenStackControlplane we'll just have to have conditionals to call the correct reconcile logic across the two of them. Also, I currently have the reconcile code guarded by targetVersion. If a targetVersion has already been reconciled, or is just not set then nothing will happen.

version_updates.md
```diff
@@ -0,0 +1,120 @@
...
+TBD
+
+## Custom images for other OpenStack services
+
+In general administrators should not need to customize containerImages
+for any other OpenStack services images. The exception would be
+the need to apply a hotfix. In order to accomidate a hotfixed or
+patched containerImage you can update the 'serviceExcludes'
+array on the OpenStackVersion to contain the camel cased
+name for the services that needs to be excluded from
```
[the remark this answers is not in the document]
dprince: The former: 'GlanceApi' was my intention. I think we could probably implement a comparator that allows a fairly sloppy interpretation of these strings too if we want. I also think having the docs for the CRD provide examples of this is probably useful.

version_updates.md
dprince: Right now any change would be a minor update. In the future for major upgrades (should we choose to reuse this resource for that workflow) we could have a SemVer check so that major changes go to one reconcile function and minor another

gibizer: Sorry I think I asked the wrong question. Do we support CRD changes during the minor upgrade solution proposed here?

version_updates.md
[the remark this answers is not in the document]
gibizer: > Container images are never really deleted however if you install a new set of openstack operator's (the entire set) it is very likely that all new image defaults will exist. The old image references will persist on any existing/deployed OpenStackControlplane and will not change or be deleted in anyway until a user kicks off the minor update process. In summary, installing new operators via OLM gives you new image defaults, and the option to perform a minor update as well. 

I'm trying to understand how the container images (like nova-api image) is managed in an offline installation. Does the old container images only manually deleted from the cluster after the OpenStackControlPlane is changed to refer only to the new images?

> As for manipulations only the image versions are getting patched via OpenStackVersion. Once we make a meaningful CRD change that requires a new version of OpenStackControlplane we'll just have to have conditionals to call the correct reconcile logic across the two of them. Also, I currently have the reconcile code guarded by targetVersion. If a targetVersion has already been reconciled, or is just not set then nothing will happen.

If I understand your answer correctly then. We allow the human to change OpenStackControlplane after a new set of operators are deployed via OLM but before the image version are changed via OpenStackVersion.  The reconciliation of such human change on OpenStackControlPlane CR will be done by the new operators but still deploying the old container images. I see the following problematic scenario.
1. greenfield deployed with a nova container version X and nova-operator version A
2. a new nova container version Y is built which contains a nova upstream change that require a tweak in the nova configuration generated by the nova-operator
3. a nova-operator version B is built that supports nova container version Y
4. a minor update is started in the deployment. The nova-operator version B is deployed by OLM but the nova container version Y is not used yet as OpenStackVersion.TargetVersion is not bumped to the new AvailableVersion
5. the human do a reconfiguration of OpenStackControlPlane CR, unrelated to the minor update, that results in the deployments of new nova containers.  As the reconciliation of such change already handled by the new nova-operator version B but still using the old nova container image X the nova-operator needs not to do the config tweak as that is only needed by nova container image Y not X. 
Do you also think this is a valid scenario? If so how should nova-operator handle this? Does nova-operator can look at the OpenStackVersion CR to see that the TargetVersion is still not bumped and therefore skip the logic that only required by the new TargetVersion?

version_updates.md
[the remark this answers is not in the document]
dprince: So it actually isn't the CRD name in some cases as we have more containerImage names than CRDs I think. It is the RELATED_IMAGE name, remove the underscores, and then CamelCase it. We could just use the RELATED_IMAGE... name verbatim but I think that wouldn't be ideal.

version_updates.md
dprince: I don't see why not so long as those changes don't modify the containerImage's themselves.

version_updates.md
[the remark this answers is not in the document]
dprince: > I'm trying to understand how the container images (like nova-api image) is managed in an offline installation. Does the old container images only manually deleted from the cluster after the OpenStackControlPlane is changed to refer only to the new images?

In an offline installation container images get mirrored like this: https://docs.openshift.com/container-platform/4.14/installing/disconnected_install/installing-mirroring-installation-images.html
Nothing that is happening here changes that. When you install a new set of operators (each time) you'd need to mirror things (operators, bundles, and all of the associated images RELATED_IMAGES...).

> If I understand your answer correctly then. We allow the human to change OpenStackControlplane after a new set of operators are deployed via OLM but before the image version are changed via OpenStackVersion. The reconciliation of such human change on OpenStackControlPlane CR will be done by the new operators but still deploying the old container images. I see the following problematic scenario.
> 
> 1. greenfield deployed with a nova container version X and nova-operator version A
> 2. a new nova container version Y is built which contains a nova upstream change that require a tweak in the nova configuration generated by the nova-operator
> 3. a nova-operator version B is built that supports nova container version Y
> 4. a minor update is started in the deployment. The nova-operator version B is deployed by OLM but the nova container version Y is not used yet as OpenStackVersion.TargetVersion is not bumped to the new AvailableVersion
> 5. the human do a reconfiguration of OpenStackControlPlane CR, unrelated to the minor update, that results in the deployments of new nova containers.  As the reconciliation of such change already handled by the new nova-operator version B but still using the old nova container image X the nova-operator needs not to do the config tweak as that is only needed by nova container image Y not X.
>    Do you also think this is a valid scenario? If so how should nova-operator handle this? Does nova-operator can look at the OpenStackVersion CR to see that the TargetVersion is still not bumped and therefore skip the logic that only required by the new TargetVersion?

If we need logic to handle complex upgrade scenarios like this I think checking the OpenStackVersion CR's deployed targetVersion and "gating" or adding a conditional around new service operator code could be a valid pattern. Is the example you give here more likely to occur during an upgrade though (instead of a minor update where no new configuration changes exist)?

version_updates.md
```diff
@@ -0,0 +1,120 @@
...
+  labels:
+  name: openstack
+spec:
+  openstackControlPlaneName: openstack
+  targetVersion: 1.0.1
+  cinderVolumeExcludes:
+    -netapp
+```
+
+## Custom Images for Neutron
```
booxter, line 89: I understand that you are targeting updates process specifically; but this new CR may be helpful in third party neutron ml2 driver integration. Let me know if this is out of scope.

---

To integrate a custom neutron ml2 driver, one would build a custom image with the driver python package (and any other dependencies). Then this custom image should replace the default one. How do we do it?  Are we allowed to override RELATED_IMAGE_..._DEFAULT for a particular deployment? Should there be serviceOverrides in addition to serviceExcludes?

version_updates.md
```diff
@@ -0,0 +1,120 @@
...
+
+TBD
```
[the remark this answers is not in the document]
dprince: Sounds good for reusing serviceExcludes for any customization to neutronApi.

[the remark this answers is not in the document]
dprince: My initial version of OpenStackVersion had images managed directly on the CR. I moved away from it as a simplification, and based on some feedback for ad-hoc hotfixes. Given the OVN orchestration requirements discussed on slack I think we have no choice but to go back to it which sounds like it aligns with your preferences as well.

version_updates.md
dprince: Need to update the docs today to reflect recent feedback. But custom ML2 drivers would get handled the same way we do custom Cinder volume backends. The administrator would edit/patch/update the OpenStackVersion spec so that a custom ML2 driver containerImage would get deployed instead of the default.

While an administrator could override RELATED_IMAGE_.. a change like that would be non-persistent on upgrades to the operator. Also, it would involve editing RELATED_IMAGE... values in multiple places (for different operator deployments). I think keeping the customizedImages in a CR might be cleaner.

version_updates.md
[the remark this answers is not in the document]
gibizer: > In an offline installation container images get mirrored like this: https://docs.openshift.com/container-platform/4.14/installing/disconnected_install/installing-mirroring-installation-images.html
> Nothing that is happening here changes that. When you install a new set of operators (each time) you'd need to mirror things (operators, bundles, and all of the associated images RELATED_IMAGES...).

Thanks! I guess similarly the cleanup of the old unused container images done manually.

> If we need logic to handle complex upgrade scenarios like this I think checking the OpenStackVersion CR's deployed targetVersion and "gating" or adding a conditional around new service operator code could be a valid pattern. Is the example you give here more likely to occur during an upgrade though (instead of a minor update where no new configuration changes exist)?

OK. I think we are on the same page here about the OpenStackVersion usage. You are right that we should not do such complex changes in a minor update.

version_updates.md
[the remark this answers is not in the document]
gibizer: good point of not always having a 1:1 mapping between CRD and container image. If we can document the list of possible names (or just document the cli command how to get is from the csv) then I'm OK with the naming scheme.

version_updates.md
gibizer: ack

gibizer (commented):
Looks good to me.

version_updates.md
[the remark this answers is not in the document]
booxter: With "Custom images for other OpenStack services" section, I don't see the need to have a separate section for Neutron.

version_updates.md
```diff
@@ -0,0 +1,130 @@
...
+  openstackControlPlaneName: openstack
+  targetVersion: 1.0.1
+  serviceCustomContainerImages:
+    GlanceApi: <custom image URL location>
+```
+
+## Custom ordering of service updates
+
+Container image updates currently occur in parallel for a given CR.
+First OVN will be updated on the Dataplane CRs. Then the Controlplane will be updated.
```
booxter, line 118: note that ovn-controller is running on both OCP workers and EDP nodes, and these should be updated before northd/ovndbclusters (both running on OCP workers only). So the first step for OVN should include both Dataplane CR update *and* ovnController section update in ControlPlane CR.

version_updates.md
```diff
@@ -0,0 +1,130 @@
...
+Finally the rest of the Dataplane CRs container images will be updated.
+
+If a situation arises where we need to update one service before
```
booxter, line 121: I think the situation just arised (see ovnController remark).

version_updates.md
```diff
@@ -0,0 +1,130 @@
...
+## Custom images for Cinder Volume
+
+In order to allow administrators to configure customized Cinder volume backends
+with custom containerImages for each backend you must specify those images
+from automatic updates. This is because there is no way to fully automate
+the creation of those images at this time. To configure a custom
+Cinder volume named 'netapp' updates you can add the name 'netapp' to the
+'cinderVolumeCustomImages' map on the OpenStackVersion CR and configure
+a custom image for that volume backend. Multiple named Cinder volume backends may
+be configured in this manner.
```
Akrog, line 70: There are a couple of things that I would like to update in this text, instead of talking about custom containerImages these would be certified vendor containerImages, also the "specify those images from automatic updates" sounds weird to me, then we talk about creation of images which can be misleading...

I think a reworked paragraph could work better in the line of:

```
In order to allow administrators to configure customized Cinder volume backends
with certified 3rd party vendor containerImages for any of the backends, 
administrators will be allowed to explicitly specify those images overriding the
default containerImage during automatic updates. This is a manual operation 
because there is no way to fully automate the selection of those images at this
time. To configure a Cinder volume backend named 'pure' with a different image 
during the updates, the administrator can add the name 'netapp' to the
'cinderVolumeCustomImages' map on the OpenStackVersion CR and set a custom image
to be used for that volume backend. Multiple named Cinder volume backends may
be configured in this manner.
```

version_updates.md
Akrog: For Cinder we currently set the `containerImage` manually in the `OpenStackControlPlane` for the specific cinder volume backend: https://github.com/openstack-k8s-operators/cinder-operator/blob/5e8e748d6046beee57bf3e4271565ff3bb7f8d11/config/samples/backends/hpe/3par/iscsi/backend.yaml#L24

version_updates.md
```diff
@@ -0,0 +1,130 @@
...
+   of OpenStackVersion.
+   
+
+## General Update Workflow
+
+OpenStack is installed by following the normal installation procedure. When
+an OpenStackControlplane resource is created in any given namespace an
+OpenStackVersion resource will be immediately created when the OpenStackControlPlane
+is reconciled (1 OpenStackVersion per namespace).
+
```
Akrog, line 33: For Cinder volume I think at this point the `cinderVolumeCustomImages` section of the `OpenStackVersion` should be automatically populated if the `containerImage` has been configured for a cinder backend.

That way the `OpenStackVersion` and `OpenStackControlplane` are in sync.

Don't know if it's worth doing it for other services. I think it may be too much.

version_updates.md
```diff
@@ -0,0 +1,130 @@
...
+In order to apply an update an administrator sets the 'TargetVersion' to the
+same value as the 'AvailableVersion' on the OpenStackVersion resources. This
+triggers the OpenStackVersion to reconcile and will update the images
+on the associated OpenStackControlplane and OpenStackDataplane resources in
+that namespace.
+
+```shell
+$ oc get openstackversion
+NAME              TARGET VERSION   AVAILABLE VERSION
+openstack         1.0.0            1.0.1
```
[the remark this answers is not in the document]
ciecierski: Target version and available version is going to be the same, no ? Or else available version is going to be empty.

version_updates.md
dprince: I added a section to describe the way it works now: if the backend does not require a custom container the default cinder
volume container is used automatically. This means if you are only using
backends that require the default cinder volume container no
edits to the OpenStackVersion resource are needed

version_updates.md
```diff
@@ -0,0 +1,130 @@
...
+is reconciled (1 OpenStackVersion per namespace).
+
+When a new set of Operators is installed via OLM (either manually or automatically)
+the OpenStackVersion resources will reconcile and a new 'AvailableVersion' will
+be set on each CR.
+
+In order to apply an update an administrator sets the 'TargetVersion' to the
+same value as the 'AvailableVersion' on the OpenStackVersion resources. This
+triggers the OpenStackVersion to reconcile and will update the images
+on the associated OpenStackControlplane and OpenStackDataplane resources in
```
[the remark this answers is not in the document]
dprince: done in the latest push

version_updates.md
```diff
@@ -0,0 +1,130 @@
...
+
+ * Use of OLM (Operator Lifecycle Manager) is assumed
+ * The CSV (ClusterServiceVersion) for the openstack-operator
+   is used as the primary resource for versioning and the
+   associated ENV variables for OpenStack service containers.
+   These are typically prefixed with RELATED\_IMAGE...
+   The CSV defines the version.
+ * OpenStack operators are always upgraded first via OLM. This
+   can be either via manual or automatic. Upgrading operators
+   will trigger OpenStackVersion to provide access to a new
```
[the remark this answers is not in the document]
dprince: fixed in the latest push

version_updates.md
dprince: I will update the docs to better word this. While there are 2 conditions for OVN Controllers (one for dataplane and one for controlplane) the intent is that they can occur independently. See here: https://github.com/openstack-k8s-operators/openstack-operator/blob/main/controllers/core/openstackversion_controller.go#L162

Then initial PR for OpenStackVersion only had the controlplane parts implemented. Now that we have those we can proceed with implementing things on the Dataplane side too.

version_updates.md
dprince: It is well understood that both the OVN controllers on the dataplane and controlplane need to be updated for our initial minor updates support. This comment was really about future unforeseen situations that may arise here. I hope this makes sense...

version_updates.md
dprince: @booxter I added expanded the section in this doc about Neutron. Still a TBD I think. There is currently no formal interface to specify custom ML2 drivers other than OVN (correct me if I am wrong here). So I think before we can add the ML2 section to this CR we should first define where and how those images would be used. As far as I know every other image we have listed on OpenStackVersion satisfies these conditions.

version_updates.md
Akrog: NetApp doesn't require a custom container, so it's misleading.  I would use Pure as an example since it does need a custom container.

version_updates.md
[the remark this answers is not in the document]
Akrog: Reading again my comment I see that it was hard to read (it hurts my eyes).
It's just that when I change the target version in order to do the upgrade, then running `oc get openstackversion` won't let us know if the update has completed, as the output during the update as well as once the update has completed would look the same:

```
$ oc get openstackversion
NAME              TARGET VERSION   AVAILABLE VERSION
openstack         1.0.1            1.0.1
```

version_updates.md
```diff
@@ -0,0 +1,160 @@
...
+In order to apply an update an administrator sets the 'TargetVersion' to the
+same value as the 'AvailableVersion' on the OpenStackVersion resources. This
+triggers the OpenStackVersion to reconcile and will update the images
+on the associated OpenStackControlplane and OpenStackDataplane resources in
+that namespace.
+
+```shell
+$ oc get openstackversion
+NAME              TARGET VERSION   AVAILABLE VERSION    DEPLOYED VERSION
+openstack         1.0.0            1.0.1                1.0.0
```
Akrog, line 47: Thank you for adding the deployed version field. :-)

gibizer (approved):
still looks good after the recent updates. Thanks for making this doc @dprince.

fmount (approved):

version_updates.md
```diff
@@ -0,0 +1,176 @@
...
+## ordering of service updates: normal vs minor updates
+
+Container image updates currently occur in parallel when updating
+custom images of the OpenStackVersion resource. On the ControlPlane
+this means that any image customizations on the OpenStackVersion
+resource will be applied immediately. And on the Dataplane
+those updates will be applied during the next 'deployment'.
+
+The exception to this is if targetVersion is changed thus causing a minor update
+sequence to occur. If a minor update is triggered a update special sequence is
```
Akrog, line 151: nit: an update

Akrog (approved):

PR #94, opened 2024-02-12
Fix RGW/Swift section name

fmount:
RGW is not a backend for Swift, we technically create and configure the object store service to use an external element (RGW) that exposes the same Swift API. In this scenario swift is not deployed at all via the swift-operator.


abays (approved):
/lgtm

PR #98, opened 2024-03-13
[tls] update tls doc for tls setting changes

stuggi:


tls.md
```diff
@@ -24,16 +24,46 @@ metadata:
...
+      enabled: false
     caBundleSecretName: myAdditionalCACerts
 ```
 
 Using the `caBundleSecretName` parameter a secret can be referenced containing any additional CA certificates, which should be added to the CA bundle.
 
+The openstack-operator creates CAs for its different usage.
+* rootca-public - used to issue certificates for routes and k8s services for public endpoint httpd vhost configuration
+* rootca-internal - used to issue certificates for all internal communication, like internal endpoint httpd vhost config, database, rabbitmq ..., except of the bellow special use case CAs
+* rootca-ovn - used to issue certificates for ovn/ovs
```
olliewalsh, line 39: probably should mention that it's used for mTLS to control access to the ovn/ovs database, but can get that in a follow up as will need a similar CA for libvirt

olliewalsh (approved):
/lgtm

tls.md
stuggi: yes, lets do that in a follow up. we need additional enhancements to the doc

PR #99, opened 2024-03-15
Updates to conditions.md

dprince:
Some updates to clarify that we want conditions to indicate that a given spec has been applied and the services are running with that active configuration.

### DIFF
--- a/conditions.md
+++ b/conditions.md
@@ -9,11 +9,14 @@ Description of conditions https://github.com/kubernetes/community/blob/master/co
 ## Conditions in openstack-k8s-operators podified controlplane operators
 * At the beginning of the reconciliation loop, controller adds a `Ready` condition which reflects the overall state of the service with `Status=Unknown`
 * When a controller performs a task, before it starts a step condition gets initialized to `Status=Unknown` e.g. DBReady, DBSync, Endpoint, ServiceRegistered, …
+* Conditions should always be re-initialized at the beginning of each reconcile. This is to ensure that changes to the spec
+  will reset conditions accordingly.
 * At the end of the reconcile loop the overall `Ready` condition gets set to `Status=True/False`, depending if the sub task conditions are fulfilled or not.
-* The service `Ready` condition with `Status=True` reflects that the service is configured, up and accessible.
-* If the `Ready` condition is `Status=False` or `Unknown` the service is not OK to answer requests
-  - `Status=False` indicates an Error that needs user intervention to fix it
-  - `Status=Unknown` indicates that the operator can progress without user intervention. i.e the operator is waiting for a job to complete
+* The Ready condition on our CRDs reflect that the expected state described by the Spec field of the CRD is the same as the actual state of the system, i.e. the operator finished reconciling the state of the system according to the Spec.
+* If the `Ready` condition is `Status=False` or `Unknown` the service has not been deployed with the currently configured spec.
+  - `Status=False` indicates that a condition has not yet been met.
+* It is acceptable to set ReadyCondition=True if replicas is 0. Any checks for an active service is actually running should
+  check that both replicas is >0 and ReadyCondition=True.
 
 ## Controllers and conditions
 General behavior of a controller to consume/set conditions:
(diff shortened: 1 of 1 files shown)


conditions.md
```diff
@@ -9,11 +9,15 @@ Description of conditions https://github.com/kubernetes/community/blob/master/co
...
 * At the beginning of the reconciliation loop, controller adds a `Ready` condition which reflects the overall state of the service with `Status=Unknown`
 * When a controller performs a task, before it starts a step condition gets initialized to `Status=Unknown` e.g. DBReady, DBSync, Endpoint, ServiceRegistered, …
+* Conditions should always be re-initialized at the beginning of each reconcile. This is to ensure that changes to the spec
+  will reset conditions accordingly.
 * At the end of the reconcile loop the overall `Ready` condition gets set to `Status=True/False`, depending if the sub task conditions are fulfilled or not.
-* The service `Ready` condition with `Status=True` reflects that the service is configured, up and accessible.
-* If the `Ready` condition is `Status=False` or `Unknown` the service is not OK to answer requests
-  - `Status=False` indicates an Error that needs user intervention to fix it
-  - `Status=Unknown` indicates that the operator can progress without user intervention. i.e the operator is waiting for a job to complete
+* The `Ready` condition with `Status=True` reflects that the service is configured, up and accessible and deployed with
```
gibizer, line 15: We would word this in a more generic way as not all of our CRDs describe a single service.
"
The Ready condition on our CRDs reflect that the expected state described by the Spec field of the CRD is the same as the actual state of the system, i.e. the operator finished reconciling the state of the system according to the Spec.
"

fmount:
+1 from me, I think we can merge as long as the wording related to @gibizer's commend is sorted out

conditions.md
dprince: works for me

gibizer (approved):

fmount (approved):
/lgtm

conditions.md
```diff
@@ -9,11 +9,14 @@ Description of conditions https://github.com/kubernetes/community/blob/master/co
...
+* Conditions should always be re-initialized at the beginning of each reconcile. This is to ensure that changes to the spec
+  will reset conditions accordingly.
 * At the end of the reconcile loop the overall `Ready` condition gets set to `Status=True/False`, depending if the sub task conditions are fulfilled or not.
-* The service `Ready` condition with `Status=True` reflects that the service is configured, up and accessible.
-* If the `Ready` condition is `Status=False` or `Unknown` the service is not OK to answer requests
-  - `Status=False` indicates an Error that needs user intervention to fix it
-  - `Status=Unknown` indicates that the operator can progress without user intervention. i.e the operator is waiting for a job to complete
+* The Ready condition on our CRDs reflect that the expected state described by the Spec field of the CRD is the same as the actual state of the system, i.e. the operator finished reconciling the state of the system according to the Spec.
+* If the `Ready` condition is `Status=False` or `Unknown` the service has not been deployed with the currently configured spec.
+  - `Status=False` indicates that a condition has not yet been met.
```
gibizer, line 17: I would keep the distinction between Unknown and False as False means error, Unknown means not finished yet to keep ourselves compatible the the k8s definitions of these values.

gibizer (changes requested):

conditions.md
fmount: agree on this one, it's good to keep the distinction between `Status=False` and `Status=Unknown`.
IIUC, if the status is `False`, there's an error, and the human operator (or an external entity) should change something in the service CR to trigger a new reconcile that can progress to a positive Status.
`Unknown`, on the other hand, feels like a not configured resource or something still in progress, but that is supposed to converge at some point without any external entity intervention.
If the above is correct ^, I think it's worth mentioning that we have a clear distinction between the two conditions.
I'm reviewing those in the Glance operator (as part of the `Conditions` update) and I'd like to be consistent with this doc, so +1 to clarify further (or just keep the previous distinction that is not wrong).

conditions.md
```diff
@@ -9,11 +9,14 @@ Description of conditions https://github.com/kubernetes/community/blob/master/co
 ## Conditions in openstack-k8s-operators podified controlplane operators
 * At the beginning of the reconciliation loop, controller adds a `Ready` condition which reflects the overall state of the service with `Status=Unknown`
 * When a controller performs a task, before it starts a step condition gets initialized to `Status=Unknown` e.g. DBReady, DBSync, Endpoint, ServiceRegistered, …
+* Conditions should always be re-initialized at the beginning of each reconcile. This is to ensure that changes to the spec
+  will reset conditions accordingly.
```
bogdando, line 13: please note about not changing transition times if the condition value doesn't change

conditions.md
Akrog: I'm looking at the [k8s API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md) and I cannot find where it states that `Ready=False` should only happen on an error.

I read this though:
```
An example of an oscillating condition type is Ready, which indicates the
object was believed to be fully operational at the time it was last probed.
```

That would indicate that on a reconciliation loop we would set conditions to `Unknown`, and once we reach the specific part of the loop for that condition we would set it to `True` if no changes are necessary and if changes are necessary they would be made and the condition set to `False` until they are in effect, at which point the condition would go to being `True`.

On the other hand, if we decide that we are going to use `False` to signal an error, I believe this cannot only be a fatal error, since we would also want to signal errors that are preventing us from advancing, such as the API returning an error.

gibizer: The background discussion: https://redhat-internal.slack.com/archives/CQXJFGMK6/p1713199689950509 convinced me to drop that we don't need to code the fatalness of an error into the condition status itself. If I want I can word it into the message (or if we want a machine interface we can word it into the reason field). So I rest my case and accept that False can mean any kind of non-readyness both automatically healed and terminal one. And Unknown simply means the controller had no way to decide as it could not run the relevant code (yet).

gibizer:
/lgtm 

Can we land it?

gibizer (approved):

PR #100, opened 2024-03-25
Update Ceph RBD Glance Image Conversion

fultonj:
Now that the Glance Operator configures this automatically we do not have to tell the user to configure it with extra steps.

### DIFF
--- a/ceph.md
+++ b/ceph.md
@@ -193,24 +193,9 @@ to access the Ceph client files, they are stored in
 
 ## Configure Glance
 
-When Glance is configured with Ceph it's recommended to enable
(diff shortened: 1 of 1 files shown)


ceph.md
```diff
@@ -233,12 +218,16 @@ spec:
         store_description = "RBD backend"
         rbd_store_pool = images
         rbd_store_user = openstack
-        [image_import_opts]
-        image_import_plugins = ['image_conversion']
-        [image_conversion]
-        output_format = raw
```
fultonj, line 239: This is now handled by the operator if necessary.

https://github.com/openstack-k8s-operators/glance-operator/blob/89a26154405ffd020ac05c761fdc6ebf4c2d3775/templates/common/config/00-config.conf#L118-L125

fmount (approved):
/lgtm

PR #102, opened 2024-04-03
Document ObservedGeneration usage

gibizer:


observed_generation.md
```diff
@@ -0,0 +1,74 @@
...
+        // subCR is successfully reconciled the latest Spec update
+}
+```
+
+## Implementation
+The `Generation` field is implemented automatically.
+
+The `Status.ObservedGeneration` needs to be implemented by every controller
+individually. This field needs to be bumped to the value of the `Generation`
+field at the start of each `Reconcile` run, after the `Status.Conditions`
```
fmount, line 46: This statement is True under the assumption that:
1. we are considering transient states a bug or just wrong
2. we are creating a strong relation between `observedGeneration` and the readiness

We can still decouple the two things, but I suspect it's a matter of having an agreement on the pattern.

dprince:
Do we really need to document observedGeneration. It seems like most of the docs  here are actually related to checking for Readyness. The pattern itself is pretty well known and the comments in the CRD should be mostly self explanatory right?

observed_generation.md
gibizer: > we are considering transient states a bug or just wrong

if the transient Ready state can cause the client to think that the last spec generation is reconciled while it isn't, then I consider that as a bug. However if along with Ready state there is an observedGeneration in the status that show what this readiness is actually about an older generation than the latest Spec generation then the client has enough input to ignore the Ready state and then it is not a bug as the Status is self consistent.

> we are creating a strong relation between observedGeneration and the readiness

From the subCR controller perspective the ReadyCondition (and all other Status fields) are calculated for a given Spec generation, and this generation value is stored in the observedGeneration. I don't see how this coupling can be avoided and still present a consistent Status field.

fmount: Yeah, I think we're on the same page for this particular point. The client will check the && condition between the Readiness and the ObservedGeneration, which makes sense when we state that: "A reconciler looks for the readiness associated to the last updated CR".
To make this happen, we need the `observedGeneration` field in the subCRs and it can be bumped up as long as a reconciling is triggered.

fmount: ah, yes, transient state can be considered as a bug, I agree on that.

fmount:
In general I think we're on the same page on the topic, we agree that there are cases where this pattern should be extended to the sub-level CRs and the top-level should compute its readiness according to the last version of the subCRs.

observed_generation.md
```diff
@@ -0,0 +1,74 @@
...
+`Spec` changes are applied to the running system, then the client needs to look
+at two fields in the CR `Status`:
+* If the `Status.ObservedGeneration == Generation` then the client can be sure
+that the controller seen the last `Spec` version
+* If the `Ready` condition's State is `True` in the `Status.Conditions` list
+then the  client can be sure that the last reconciliation finished
+successfully.
+
+```golang
+if subCR.Generation == subCR.Status.ObservedGeneration &&
```
fmount, line 35: wondering if as per @mrkisaolamb this should be: `subCR.Generation <= subCR.Status.ObservedGeneration`, and should be updated when we rollout the new pattern

gibizer: that is a safe change as far as I understand, however I would like to have a clear reason to make such change. I know that you also asked for reasons from @mrkisaolamb as well.

observed_generation.md
```diff
@@ -0,0 +1,74 @@
...
+of the parent CR depends on the `subCR.Status` fields (mostly the
+`Status.Conditions`) and therefore it needs to consider what `Spec`
+`Generation` those `Status` fields are referring to. So the
+`subCR.Status.ObservedGeneration` needs to be included in the parent CR's
+Readiness calculation. This is needed as the subCR's `Ready` condition alone
+cannot tell if the controller of the subCR see the latest `Spec` update from
+the controller of the parent CR.
+
+## When a CR depends on Pods
+We have CRs that are managing `Pods` via `Deployment`, `StatefulSet` or other
```
Akrog, line 65: Cinder and Glance operators are not doing this right.
I'm fixing it in the cinder operator while I add the observedGeneration code.

Akrog (approved):
Thank you for this doc, it is really clarifying and helpful.

I only have a nit on a conjugation.

observed_generation.md
fmount: you mean that we should set `ReadyCount` if the `observedGeneration` is the last seen?

Akrog: Yes, because right now this is being done unconditionally (at least in glance and cinder).

Example: https://github.com/openstack-k8s-operators/glance-operator/blob/6b5994d3fe9453fbaad2fff857b152cd75db7d44/controllers/glanceapi_controller.go#L780

fmount: Yeah, and https://github.com/openstack-k8s-operators/glance-operator/pull/508 cover the glance part of it

fmount: I did a similar change for manila: https://github.com/openstack-k8s-operators/manila-operator/pull/264/commits/9ec049c88df58489ecc7f913f3f55f0a85a85342

fmount (approved):
/lgtm

observed_generation.md
gibizer: OK sounds like we are on the same page

observed_generation.md
fmount: I think it won't hurt for those edge cases where something really wrong might happen. See also [1].
I'm inclined to rollout this pattern, because 99% of the cases it would be a `==`, unless something wrong happened. I'm not even sure we will be lucky enough to see that scenario happening, but I'm not sure service operators should explicitly handle the `<` case, unless we don't want to.

[1] https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L712

mrkisaolamb: I paste here same comment and links from placement patch where we start discussion: 

> Unfortunatly IMO there is not any good reading about this. My comment is based on some comments in pr's, stack overflow which I saw and discussions and based on gh search of this pattern (eg. k8s examples above and https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/install/status_viewer.go#L13 ).
> Only docs that mention this pattern are without any additional info except that Generation <= ObservedGeneration is used there. eg https://github.com/kubernetes/design-proposals-archive/blob/acc25e14ca83dfda4f66d8cb1f1b491f26e78ffe/api-machinery/garbage-collection.md?plain=1#L207 or https://github.com/kubernetes/design-proposals-archive/blob/acc25e14ca83dfda4f66d8cb1f1b491f26e78ffe/api-machinery/garbage-collection.md?plain=1#L168 . Also there is the same question asked on k8s forum https://discuss.kubernetes.io/t/in-what-case-of-deployment-deployment-generation-deployment-status-observedgeneration-is-true/24655 without answers

gibizer: I can add a sentence to the doc. I think I see the case where generation legitimately <= than observedGeneration:
1. our controller updates a subCR.Spec and reads back the result. It sees generation N
2. other client updates the subCR.Spec, generation is now N+1
3. the subCR controller reconciles the Spec version N+1 and sets observedGeneration to N+1
4. our controller reads the subCR.Status from the API  (there is a status subresource so it can be read independently) and sees observedGeneration N+1 while it still sees generation as N

However I think in this case our controller cannot decide if the subCR.Spec is according to its needs, maybe the other client set something on the spec that our controller needs to reset. So I think in this case our controller should not state that the subCR is up to date but instead requeue a reconcile that will update the subCR.Spec again if needed.

fmount: I'm not sure we need to explicitly handle the `<` case and return. In theory we can continue the reconciliation loop if `<=`, and not stop. The statefulset controller is supposed to handle any inconsistencies before it returns Ready to our controller.

gibizer: in the scenario I described above `<` means that the controller sees the status of a newer Spec generation that is not set by the controller in the current reconcile. I think in this case the controller cannot say that the underlying resource is ready according to the controllers request because something else made and additional request since then and the controller does not know if that additonal request is compatible with the spec it wants to apply on the underlying resource.

fmount: So, I think that `==` is sufficient to continue the reconciliation loop, because `<` would be covered by such condition, we don't have to explicitly return, we are just not able to reach the readiness.

gibizer: I agree. This is why I kept the example in the doc using == when I added the explanation about a possible case of <

fmount: Just to close this thread. I think by having `==` instead of `<=` we're refusing the `DeploymentComplete` status provided by [1], and have an additional reconciliation loop, which I think it was what we envisioned at the beginning. 
If we use `<=` I think a new reconciliation would happen anyway (based on the logic I see in most of the operators), but I I like more finding an agreement on the pattern rather than continue the `<=` investigation, based on theories we're not often able to clearly reproduce (it would require time and effort).

[1] https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L712

fmount:
@gibizer @mrkisaolamb @Akrog @dprince I think we might want to use `==` in the `Generation` / `ObservedGeneration` comparison (and eventually explicitly handle the `<` case, where something wrong happened), so I think we can merge this patch if there are no additional concerns.

observed_generation.md
gibizer: > If we use <= I think a new reconciliation would happen anyway 
yes, but the controller might set ready true in the current reconcile already causing a too early ready state

PR #112, opened 2024-05-22
Add guide about testing locally openstack code changes with podified env

slawqo:
This patch adds guide about how to easily test locally changes in the OpenStack code on the podified environment. This method can also be used for the development workflow, not only testing.

This document is based on the similar guide from the edpm-ansible repo [1] which is about testing edpm-ansible roles with ansibleee operator.

[1] https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/docs/source/testing_with_ansibleee.rst

Related: #OSP-26653

### DIFF
--- a/testing_openstack_code_changes_with_operators.md
+++ b/testing_openstack_code_changes_with_operators.md
@@ -1,0 +1,126 @@
+# Testing OpenStack code changes in the local podified environment
+
+You can test your changes made to the OpenStack project (Neutron will be used as
+an example in this document) in the OpenStack deployed on top of the OpenShift,
+like e.g. CRC. To do so, couple of steps are needed.
+This guide is based on the similar guide from the
+[edpm-ansible](https://github.com/openstack-k8s-operators/edpm-ansible/blob/main/docs/source/testing_with_ansibleee.rst)
+repository.
+
+## Create local NFS storage
+
+When using OpenShift Local (aka CRC), your export will be something like this:
+
+```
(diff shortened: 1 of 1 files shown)


testing_openstack_code_changes_with_operators.md
```diff
@@ -0,0 +1,122 @@
...
+    - ReadOnlyMany
+  resources:
+    requests:
+      storage: 1Gi
+EOF
+
+oc apply -f openstack-code-storage.yaml
+```
+
+## Add extraMount to Your OpenStackControlPlane CR
```
stuggi, line 82: maybe there should be a node that this procedure only work with services supporting `extraMounts`. from a quick look at the ctlplane crd its about 5 or 6 from ~20.

slawqo: good point, done :)

testing_openstack_code_changes_with_operators.md
```diff
@@ -0,0 +1,126 @@
...
+Make sure nfs-server and firewalld are started:
+```
+% systemctl start firewalld
+% systemctl start nfs-server
+```
+
+Tip
+---
+CRC installs its own firewall rules, which likely will need to be adjusted
+depending on the location of your NFS server. If your edpm-ansible directory is
```
karelyatin, line 29: s/edpm-ansible/neutron

karelyatin (commented):
/lgtm

just a typo inline

testing_openstack_code_changes_with_operators.md
slawqo: fixed, thx

fmount:
awesome, thanks @slawqo /lgtm as well

PR #115, opened 2024-06-07
Additional Subject Alternative Names to TLS doc section

fmount:
This patch describes the implementation of a mechanism to add `Subject Alternative Name` `DNS` entries to the `Certificate` Requests.
An operator can provide to the `Service` the `Annotation` described by this patch to get the additional entries processed by the `openstack-operator`.

Jira: https://issues.redhat.com/browse/OSPRH-7276


stuggi (approved):
/lgtm

PR #116, opened 2024-06-11
Improve extraMounts docs with samples and a few additional details

fmount:
This patch improves the `extraMounts` description with more details that might be useful to produce documentation.

Related: https://issues.redhat.com/browse/RHOSPDOC-1627


fultonj (approved):
thanks Fracnesco

PR #124, opened 2024-08-07
Troubleshoot Ceph network connection without nc

fultonj:
The `nc` command is not installed on the Glance pod so we cannot suggest it be used to test the pod's connection to Ceph. Instead the procedure is re-written using a Python socket since that has confirmed to be available in the Glance pod.

### DIFF
--- a/troubleshooting.md
+++ b/troubleshooting.md
@@ -149,17 +149,45 @@ the [networking document](networking.md). The IPs and ports of the Ceph
 cluster's monitors should be in `/etc/ceph.conf`. From within the pod, read
 this and try to reach the IP as well as connect to the port.
 ```
-$ oc debug --container glance-api pod/glance-default-external-api-0
-bash-5.1$ cat /etc/ceph/ceph.conf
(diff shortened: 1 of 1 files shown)


troubleshooting.md
```diff
@@ -149,17 +149,45 @@ the [networking document](networking.md). The IPs and ports of the Ceph
 cluster's monitors should be in `/etc/ceph.conf`. From within the pod, read
 this and try to reach the IP as well as connect to the port.
 ```
-$ oc debug --container glance-api pod/glance-default-external-api-0
-bash-5.1$ cat /etc/ceph/ceph.conf
-# minimal ceph.conf for 7fbaf2fd-f70e-5625-b00e-65b33ee823a6
+$ oc get pods | grep glance | grep external-api-0
```
fmount, line 152: I think we can remove `| grep glance` as we're going to get the same `Pod` in the example

fmount (commented):
small nit, looks good otherwise!

troubleshooting.md
fultonj: done

fmount (approved):
/lgtm

PR #125, opened 2024-08-09
Publish an OpenStack testing image to openshift-image-registry

fmount:
This patch improves the existing guide to give an alternative to the `quay.io` usage.
It is possible to import the image (built with `podman` or `tcib`) within the existing `openshift-image-registry`, and consume it using the regular `OpenStackVersion` mechanism.


fmount:
I'd like to turn this doc into a generic procedure (build an `openstack image container` and test it), but I'll do that in a follow up.

PR #128, opened 2024-10-16
Improve patch command for removing webhook defs from CSV

abays:
Rather than outright removing `/spec/webhookdefinitions` from a CSV, we should instead document replacing them with an empty array.  This is preferable because it can be executed again without returning an error, unlike the removal which fails if the path does not exist.


fmount (approved):
/lgtm

PR #137, opened 2025-03-21
Add edpm update overview docs

slagle:
Depends-On: https://github.com/openstack-k8s-operators/edpm-ansible/pull/911
Signed-off-by: James Slagle <james.slagle@gmail.com>

### DIFF
--- a/edpm_update_overview.md
+++ b/edpm_update_overview.md
@@ -1,0 +1,46 @@
+# Overview of Updating Services on EDPM Nodes
+
+This document gives a high level overview of updating services on RHEL-based
+External Dataplane Management (EDPM) nodes.
+
+## The update service
+
+The EDPM update process is typically executed in two separate steps of updating
+ovn, then updating the rest of the services. For more details on this process
+see [Updating the data
+plane](https://openstack-k8s-operators.github.io/openstack-operator/dataplane/#assembly_updating-the-data-plane).
+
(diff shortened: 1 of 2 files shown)


edpm_update_overview.md
```diff
@@ -0,0 +1,46 @@
+# Overview of Updating Services on EDPM Nodes
+
+This document gives a high level overview of updating services on RHEL-based
+External Dataplane Management (EDPM) nodes.
+
+## The update service
+
+The EDPM update process is typically executed in two separate steps of updating
+ovn, then updating the rest of the services. For more details on this process
```
bogdando, line 9: Please note about the nova compute vs control plane Nova update ordering
According to https://issues.redhat.com/browse/OSPRH-13923
```
The nova-compute services on the EDPM node should only be updated after the full control plane update is finished. (I.e. be careful when updating the OVN services on the EDPM node not to include the nova service to that step)
```

slagle: these aren't meant to be user docs. the only reason I mention ovn here is because what I link to describes ovn update ordering. What you want me to add sounds like nova update docs for the user, in which case they should be added elsewhere.

I don't object to having that noted here though, but I think you should probably add it in a followup.

slagle:
@bshephar @gibizer @jistr Please review if you have any input. The point of this is just to document the current existing process, and not propose anything new.

PR #141, opened 2025-06-26
doc: pprof to connect to controller managers

dprince:


stuggi (commented):
/lgtm an alternative would be to run it local using the make target and use pprof against the local running one

PR #145, opened 2025-07-23
Update webhook doc

ratailor:
This change updates the webhook doc NOTE mentioning that the `oc delete` on `ValidatingWebhookConfiguration` and `MutatingWebhookConfiguration` resources created by `make run-with-webhook` command, run automatically when it receives SIGINT or SIGTERM signal.
Also add the command to scale-up the replicas of controller-manager which was scale-down with `make run-with-webhook` command.


abays (approved):
/lgtm

PR #168, opened 2026-04-15
[b/r] Add comprehensive backup/restore user guide

stuggi:
Step-by-step backup and restore procedure without Ansible dependencies. Covers prerequisites (OCP 4.20+, OADP 1.5+), backup (operator version recording, Galera dumps, OADP PVC + resources backup), restore (ordered OADP restores, staged deployment, database restore, EDPM deployment), verification, ExtraMounts PVC labeling, annotation overrides, and known limitations.

### DIFF
--- a/backup-restore/user-guide.md
+++ b/backup-restore/user-guide.md
@@ -1,0 +1,686 @@
+# OpenStack Backup and Restore User Guide
+
+This guide describes how to back up and restore an OpenStack control plane
+deployed on OpenShift using OADP (OpenShift API for Data Protection).
+
+## How It Works
+
+### Full Backup, Selective Restore
+
+The backup captures a complete snapshot of the OpenStack namespace. The restore
+is selective — only resources that operators cannot recreate are restored:
+
+| Restored | Not restored (operators rebuild) |
+|---|---|
+| User-provided Secrets (passwords, SSH keys) | Operator-managed Secrets |
+| User-provided ConfigMaps | Operator-managed ConfigMaps |
+| CA certificates | TLS service certificates (reissued) |
+| PVCs with data (Glance images, DB dumps) | Service deployments |
+| User-created CRs (ControlPlane, Version, NodeSet) | Service CRs (Nova, Keystone, Galera, ...) |
+| RabbitMQ default-user credentials | MariaDBAccount/MariaDBDatabase CRs |
+
+The restored resources serve as "seeds" — operators reconcile and recreate
+everything else from those inputs.
+
+### Labeling
+
+Resources are selected for restore using Kubernetes labels
+(`backup.openstack.org/restore`, `backup.openstack.org/restore-order`).
+Labels are applied automatically:
+
+- **OpenStackBackupConfig controller** labels CR instances and user-provided
+  Secrets, ConfigMaps, and NADs
+- **Operators** label PVCs, CA cert secrets, custom Issuers, and RabbitMQ
+  credentials at creation time
+
+No manual labeling is required for standard deployments. PVCs passed via
+`extraMounts` must be labeled manually (see [ExtraMounts PVCs](#extramounts-pvcs)).
+
+### Staged Deployment
+
+The restore uses staged deployment to ensure databases are restored before
+services start. The OpenStackControlPlane is restored with a
+`deployment-stage: infrastructure-only` annotation, which starts only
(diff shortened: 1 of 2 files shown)


backup-restore/user-guide.md
```diff
@@ -0,0 +1,671 @@
...
+
+Ensure the target namespace is clean — no stale PVCs or resources from a
+previous deployment:
+
+```bash
+# Delete all PVCs
+oc delete pvc --all -n openstack
+
+# Delete remaining resources
+oc delete all --all -n openstack
```
abays: What if there are old CRs lingering that would just regenerate the pods, deployments, statefulsets, etc?

backup-restore/user-guide.md
```diff
@@ -0,0 +1,671 @@
...
+  JOB_NAME="${CRONJOB}-${BACKUP_TS}"
+  oc -n openstack create job --from=cronjob/${CRONJOB} ${JOB_NAME} \
+    --dry-run=client -o json | \
+    jq '.spec.template.spec.containers[0].env += [{"name":"BACKUP_TIMESTAMP","value":"'${BACKUP_TS}'"}]' | \
+    oc -n openstack create -f -
+done
+
+# Wait for all backup jobs to complete
+oc -n openstack wait --for=condition=complete job -l app=galera --timeout=10m
+```
```
abays, line 134: Should we wait for `GaleraBackup` to be ready here, like we do in the CIFMW roles?  Or is this functionally equivalent?

backup-restore/user-guide.md
```diff
@@ -0,0 +1,671 @@
...
+spec:
+  includedNamespaces:
+  - openstack
+  labelSelector:
+    matchLabels:
+      backup.openstack.org/backup: "true"
+  snapshotVolumes: true
+  defaultVolumesToFsBackup: false
+  snapshotMoveData: true
+  volumeSnapshotLocations: []
```
abays: Should we provide in-line comments explaining what these options do, since they're uniquely found in this CRD?

backup-restore/user-guide.md
```diff
@@ -0,0 +1,671 @@
...
+oc get pods -n openstack --no-headers | grep -v Completed | grep -v Running
+
+# OpenStack endpoints
+oc exec -t openstackclient -- openstack endpoint list
+
+# Verify compute services are online
+oc exec -t openstackclient -- openstack compute service list
+
+# Verify network agents are up
+oc exec -t openstackclient -- openstack network agent list
```
abays: Maybe we should add `-n openstack` to these.

backup-restore/user-guide.md
stuggi: in this guide here right now we just mention it as a pre-req `5. **GaleraBackup CRs** configured for each Galera database instance`.  this wait here is then that the actual ad-hoc DB backup job finished. maybe we add something for the GaleraBackup with the downstream doc? or you think we also should cover how to configure/use GaleraBackup CR?

backup-restore/user-guide.md
stuggi: yes that probably makes sense

backup-restore/user-guide.md
stuggi: yes, I think these steps need to be better, maybe we just link to
https://docs.redhat.com/en/documentation/red_hat_openstack_services_on_openshift/18.0/html/maintaining_the_red_hat_openstack_services_on_openshift_deployment/assembly_removing-rhoso-deployment-from-rhocp-environment#proc_removing-RHOSO-resources-from-RHOCP-environment_clean  ?

backup-restore/user-guide.md
stuggi: yes, good catch. it should be consistent with the first commands

backup-restore/user-guide.md
abays: Good idea

backup-restore/user-guide.md
abays: I was misunderstanding what `GaleraBackup` does.  This is fine.

backup-restore/user-guide.md
stuggi: done

backup-restore/user-guide.md
stuggi: done, for this and other yamls. for restore I did not add it to all of them as they are the same, just the selector changes

abays (approved):
/lgtm

PR #169, opened 2026-04-16
[b/r] Move OVN sync to after EDPM in user guide, document network break

stuggi:
Move the neutron-ovn-db-sync-util step to after the EDPM deployment (Step 11) so all compute chassis are registered before flows are programmed. Document the VM network connectivity break as a known limitation caused by ovn-controller connecting to the empty SB DB.

Jira: OSPRH-22912
Jira: OSPRH-22913

### DIFF
--- a/backup-restore/user-guide.md
+++ b/backup-restore/user-guide.md
@@ -579,7 +579,21 @@ EOF
 fi
 ```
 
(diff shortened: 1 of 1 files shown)


abays (approved):
/lgtm

PR #170, opened 2026-04-17
[b/r] User guide: clarify OADP Resource Modifier ConfigMap

stuggi:
Clarify that the Resource Modifier ConfigMap is an OADP/Velero feature created in the openshift-adp namespace, referenced by Restore CRs via spec.resourceModifier.


abays (approved):
/lgtm

PR #173, opened 2026-04-28
Clarify backup step references in B/R user guide

abays:
Just adding some extra specificity to certain restore steps of the backup / restore user guide, to make it clear that they're referring to steps from the backup section (and not steps from the same restore section).


stuggi (approved):
/lgtm

PR #174, opened 2026-05-04
[b/r] Add pre-backup health check to user guide

stuggi:
Verify that control plane and data plane nodesets are Ready before taking a backup to prevent backing up a broken state.


abays (approved):
/lgtm

PR #175, opened 2026-05-11
[b/r] Backup and restore docs for OSBMS and Metal3 BMHs

abays:
First iteration.  We can follow-up with a newer version that utilizes [1] once we properly test it.  This is the current verified procedure.

Jira: https://redhat.atlassian.net/browse/OSPRH-29986

[1] https://github.com/openstack-k8s-operators/openstack-baremetal-operator/pull/394

### DIFF
--- a/backup-restore/backup/README.md
+++ b/backup-restore/backup/README.md
@@ -7,7 +7,8 @@ environments. The backup playbooks and templates are in the
 
 ## Backup Approach
 
-We use a two-backup strategy:
+We use a two-backup strategy, with an optional third backup for
+baremetal-provisioned environments:
 
 1. **PVC backup** — PVCs with CSI snapshots
    - Filters at backup time using `labelSelector`
@@ -21,7 +22,12 @@ We use a two-backup strategy:
    - Excludes PVCs (backed up separately)
    - Includes: CRs, Secrets, ConfigMaps, NADs, etc.
 
-## Why Two Backups?
+3. **BMH backup** (optional) — BaremetalHosts in a separate namespace
+   - Only needed when Metal3 BaremetalHosts are in a different namespace
+     from the OpenStack resources
+   - Includes BaremetalHosts, Secrets, and ConfigMaps
+
+## Why Separate Backups?
 
 - **Flexibility**: Allows selective restore (e.g., restore only PVCs or only CRs)
 - **PVC Filtering**: Only backup PVCs we explicitly labeled (saves storage costs)
@@ -169,7 +175,48 @@ oc wait --for=jsonpath='{.status.phase}'=Completed \
   backup/openstack-backup-resources-${BACKUP_TS} -n openshift-adp --timeout=30m
 ```
 
-### Step 5: Verify
+### Step 5: OADP Baremetal Backup (optional — baremetal-provisioned nodes only)
+
+Skip this step if all OpenStackDataPlaneNodeSets use `preProvisioned: true`.
+
+If your Metal3 BaremetalHosts are in the same namespace as the OpenStack
+resources, they are already captured by the resources backup in Step 4 and
+this step can be skipped.
+
+If your BaremetalHosts are in a different namespace, create an additional
+backup for that namespace. During restore, the BaremetalHosts, secrets, and
+configmaps from this backup are restored in Restore Step 7:
+
+```bash
+BMH_NAMESPACE=<namespace where BaremetalHosts are located>
+BMH_BACKUP=openstack-backup-bmh-${BACKUP_TS}
+
+cat <<EOF | oc apply -f -
+apiVersion: velero.io/v1
+kind: Backup
+metadata:
+  name: ${BMH_BACKUP}
+  namespace: openshift-adp
+  annotations:
+    openstack.org/csv-version: "${CSV_VERSION}"
+    openstack.org/catalog-source-image: "${CATALOG_IMG}"
+    openstack.org/operator-image: "${OPERATOR_IMG}"
+spec:
+  includedNamespaces:
+  - ${BMH_NAMESPACE}
+  includedResources:
+  - baremetalhosts.metal3.io
+  - secrets
+  - configmaps
+  storageLocation: velero-1
+  ttl: 720h
+EOF
+
+oc wait --for=jsonpath='{.status.phase}'=Completed \
+  backup/${BMH_BACKUP} -n openshift-adp --timeout=30m
+```
+
+### Step 6: Verify
 
 ```bash
 oc get backup -n openshift-adp -o custom-columns=NAME:.metadata.name,PHASE:.status.phase
@@ -195,6 +242,16 @@ All resources in the namespace except PVCs. This includes CRs, Secrets,
 ConfigMaps, NetworkAttachmentDefinitions, and any other namespace-scoped
 resources. Restore ordering is controlled by `backup.openstack.org/restore-order`
 labels on the backed-up resources (see [`../restore/README.md`](../restore/README.md)).
+
+### backup-openstack-bmh (optional — baremetal-provisioned nodes only)
+
+Metal3 BaremetalHosts and associated Secrets and ConfigMaps (Step 5). This
+backup is only needed when BaremetalHosts are in a different namespace from
+the OpenStack resources. If they are in the same namespace, they are
+captured by the resources backup above.
+
+`OpenStackBaremetalSet` and `OpenStackProvisionServer` resources are in the
+`openstack` namespace and are included in the resources backup.
 
 ## Verification
 
--- a/backup-restore/restore/README.md
+++ b/backup-restore/restore/README.md
@@ -19,6 +19,7 @@ before starting the next.
 | 30 | OpenStackControlPlane, Reservation | **Restored with `deployment-stage: infrastructure-only` annotation** |
 | 40 | GaleraBackup, IPSet, DataPlaneService | Backup config, IP sets, custom DataPlane services |
 | 50 | **Manual**: Database restore | Create GaleraRestore CRs, restore databases, remove deployment-stage annotation |
+| 55 | BaremetalHosts (Step 7), OpenStackBaremetalSet (Step 8) | Optional: baremetal-provisioned nodes only. BMH secrets/CMs restored here if separate namespace. Requires BMH pause annotation (resource modifier), operator scale-down, and `restoreStatus` |
 | 60 | OpenStackDataPlaneNodeSet | DataPlane resources (optional) |
 | - | OpenStackDataPlaneDeployment | Resync credentials on dataplane nodes |
 | - | **Manual** | Re-enable InstanceHa (`spec.disabled: False`) after verifying the cloud is operational |
@@ -129,7 +130,9 @@ for details.
 
 This ConfigMap is referenced by all Restore CRs. It strips
 `last-applied-configuration` annotations, adds the `deployment-stage`
-annotation to the ControlPlane CR, and disables InstanceHa during restore.
+annotation to the ControlPlane CR, disables InstanceHa during restore,
+and pauses any BaremetalHosts to prevent Metal3 from acting on them
+before their status is restored.
 
 ```bash
 cat <<'EOF' | oc apply -f -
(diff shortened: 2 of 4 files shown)


backup-restore/restore/README.md
```diff
@@ -352,6 +365,113 @@ oc wait openstackcontrolplane -n openstack --for=condition=Ready --timeout=30m \
...
+oc get bmh -n ${BMH_NAMESPACE}
+```
+
+### Step 6d: Restore OpenStackBaremetalSets (Order 55, optional — baremetal-provisioned only)
+
+Skip this step if all OpenStackDataPlaneNodeSets use `preProvisioned: true`.
+
+Scale down the openstack-baremetal operator and delete its webhooks.
+Velero restores spec and status in separate API calls — the validating
+webhook rejects the restore because the status is initially empty.
```
stuggi, line 460: right, we should follow up on this

backup-restore/restore/README.md
```diff
@@ -352,6 +365,113 @@ oc wait openstackcontrolplane -n openstack --for=condition=Ready --timeout=30m \
...
+  namespace: openshift-adp
+spec:
+  backupName: ${RESOURCES_BACKUP}
+  includedNamespaces:
+  - openstack
+  includedResources:
+  - openstackbaremetalsets.baremetal.openstack.org
+  - openstackprovisionservers.baremetal.openstack.org
+  - secrets
+  - configmaps
```
stuggi: as mentioned above for secrets/cms

backup-restore/user-guide.md
```diff
@@ -247,6 +247,50 @@ oc wait --for=jsonpath='{.status.phase}'=Completed \
   backup/openstack-backup-resources-${BACKUP_TS} -n openshift-adp --timeout=30m
 ```
 
+### Step 5b: OADP Baremetal Backup (optional — baremetal-provisioned nodes only)
+
+Skip this step if all OpenStackDataPlaneNodeSets use `preProvisioned: true`.
+
+If your Metal3 BaremetalHosts are in the same namespace as the OpenStack
+resources, they are already captured by the resources backup in Step 5 and
+this step can be skipped.
```
stuggi, line 256: maybe we also run it for the openstack namespace, just to make it easier for restore? see my comments on the restore side related to secrets/cm, or is there an easy way to filter them just for metal3 related secrets/cm?

backup-restore/restore/README.md
```diff
@@ -352,6 +365,113 @@ oc wait openstackcontrolplane -n openstack --for=condition=Ready --timeout=30m \
...
+  name: openstack-restore-55-bmh-${RESTORE_SUFFIX}
+  namespace: openshift-adp
+spec:
+  backupName: ${BMH_BACKUP}
+  includedNamespaces:
+  - ${BMH_NAMESPACE}
+  includedResources:
+  - baremetalhosts.metal3.io
+  - secrets
+  - configmaps
```
stuggi, line 404: if BMH_NAMESPACE ==openstack I think we have to be able to limit it to bmh only secrets and cms, otherwise we restore objects which we want to get created, like new tls certs. is there a label we can use?

* Are those user created? if so , they should not have an owner ref and got labeled to be restored in order 10
```
      backup.openstack.org/restore: "true"
      backup.openstack.org/restore-order: "10"
```
* or do they have a label that we could limit it to the ones for osbmset/prov servers?

backup-restore/user-guide.md
```diff
@@ -618,6 +674,123 @@ oc wait openstackcontrolplane -n openstack --all \
...
+  name: openstack-restore-55-bmh-${RESTORE_SUFFIX}
+  namespace: openshift-adp
+spec:
+  backupName: ${BMH_BACKUP}
+  includedNamespaces:
+  - ${BMH_NAMESPACE}
+  includedResources:
+  - baremetalhosts.metal3.io
+  - secrets
+  - configmaps
```
stuggi, line 715: same as above

backup-restore/user-guide.md
```diff
@@ -618,6 +674,123 @@ oc wait openstackcontrolplane -n openstack --all \
...
+
+oc wait --for=jsonpath='{.status.phase}'=Completed \
+  restore/openstack-restore-55-bmh-${RESTORE_SUFFIX} -n openshift-adp --timeout=10m
+```
+
+After the restore completes, remove the pause annotations so Metal3
+recognizes the BMHs in their restored state:
+
+```bash
+oc annotate bmh -n ${BMH_NAMESPACE} --all baremetalhost.metal3.io/paused-
```
stuggi, line 763: could also start removing it on one for test, before removing it on all bmh at once

backup-restore/user-guide.md
```diff
@@ -618,6 +674,123 @@ oc wait openstackcontrolplane -n openstack --all \
...
+spec:
+  backupName: ${RESOURCES_BACKUP}
+  includedNamespaces:
+  - openstack
+  includedResources:
+  - openstackbaremetalsets.baremetal.openstack.org
+  - openstackprovisionservers.baremetal.openstack.org
+  - secrets
+  - configmaps
+  restoreStatus:
```
stuggi: same as above. we can not restore just all secrets/cms in the openstack namespace from the backup, or the backup just has the bmset related secrets/cms

abays:
@stuggi I tried addressing all your comments in the force-push commit.  Please see if I've moved things forward in your opinion.

backup-restore/backup/README.md
```diff
@@ -169,6 +175,48 @@ oc wait --for=jsonpath='{.status.phase}'=Completed \
   backup/openstack-backup-resources-${BACKUP_TS} -n openshift-adp --timeout=30m
 ```
 
+### Step 4b: OADP Baremetal Backup (optional — baremetal-provisioned nodes only)
```
stuggi: maybe we just make it new optional `Step 5` instead of 4b?

backup-restore/restore/README.md
```diff
@@ -235,6 +248,38 @@ oc wait --for=jsonpath='{.status.phase}'=Completed \
   restore/openstack-restore-10-foundation-${RESTORE_SUFFIX} -n openshift-adp --timeout=5m
 ```
 
+### Step 2b: Restore BaremetalHost Secrets/ConfigMaps (Order 10, optional — separate BMH namespace only)
```
stuggi: should we make it new optional 3, could also make it a separate pre-req step using its own oadp resource modified config map and also has the bmh restore. not sure whats easier to consume.

backup-restore/restore/README.md
```diff
@@ -352,6 +397,119 @@ oc wait openstackcontrolplane -n openstack --for=condition=Ready --timeout=30m \
   $(oc get openstackcontrolplane -n openstack -o jsonpath='{.items[0].metadata.name}')
 ```
 
+### Step 6c: Restore BaremetalHosts (Order 55, optional — baremetal-provisioned only)
```
stuggi: maybe we make BMH restore its own step 7 . either with 2 sub sections, or if we move metal3 restore to its own pre section have it there.

backup-restore/user-guide.md
```diff
@@ -247,6 +247,52 @@ oc wait --for=jsonpath='{.status.phase}'=Completed \
   backup/openstack-backup-resources-${BACKUP_TS} -n openshift-adp --timeout=30m
 ```
 
+### Step 5b: OADP Baremetal Backup (optional — baremetal-provisioned nodes only)
```
stuggi: maybe we make it the new 6 ?

backup-restore/user-guide.md
```diff
@@ -359,6 +407,16 @@ data:
       - patchData: |
           spec:
             disabled: "True"
+    # Rule 4: BaremetalHosts — pause BMHs during restore so Metal3 does
```
stuggi, line 410: like above, I was wondering if bmh restore could be its own dedicated pre section and then either use its own resource modified cm. what you think about this? would it be a good split?

backup-restore/user-guide.md
```diff
@@ -618,6 +708,130 @@ oc wait openstackcontrolplane -n openstack --all \
   --for=condition=Ready --timeout=30m
 ```
 
+### Step 9b: Restore BaremetalHosts (optional — baremetal-provisioned nodes only)
```
stuggi: same as comment above, should we do a pre-section on bmh restore?

backup-restore/user-guide.md
```diff
@@ -618,6 +708,130 @@ oc wait openstackcontrolplane -n openstack --all \
...
+
+Verify BMHs are in the expected state:
+
+```bash
+oc get bmh -n ${BMH_NAMESPACE}
+```
+
+All BMHs should show `STATE: provisioned` and `ONLINE: true`.
+
+### Step 9c: Restore OpenStackBaremetalSets (optional — baremetal-provisioned nodes only)
```
stuggi: lets make it the new 10

stuggi (commented):
content looks good. I was just thinking if we should make metal3 restore its own pre section as its a restore for a "separate" product? what you think about it?

abays:
> content looks good. I was just thinking if we should make metal3 restore its own pre section as its a restore for a "separate" product? what you think about it?



> content looks good. I was just thinking if we should make metal3 restore its own pre section as its a restore for a "separate" product? what you think about it?

Sure, that makes sense.

backup-restore/backup/README.md
abays: Done

backup-restore/restore/README.md
abays: The instructions have been updated to say to skip the BMH secret/CM restore if `namespace == openstack`.  In such a case, all relevant BMH secrets/CMs, whether user-created or auto-generated, will have been restored in earlier steps.

backup-restore/user-guide.md
abays: The BMHs can effectively be restored at any time earlier than the `OpenStackBaremetalSet` restoration, so having BMHs in the `openstack` namespace get backed up with the rest of the resources there is fine, I think.  Of course, on the restore side, this assumes the user has followed instructions for the creation of the `openstack-restore-resource-modifiers` `ConfigMap`, which should include the needed pause annotation resource modified to prevent the BMH from reconciling before its status is injected by Velero.

backup-restore/user-guide.md
abays: I'd rather keep the resource modifiers all in one `ConfigMap` to simplify things.  It doesn't hurt the CM to have the extra modifiers, since they simply won't apply if there are no valid targets in the `Backup` they're modifying via a `Restore`.  But maybe it makes more sense to have an independent BMH section.  The problem is that it still won't be cleanly divided from this section, because when the BMHs are in the `openstack` namespace, they will be backed up (along with their secrets/CMs) when the `openstack` namespace resources are backed-up.  That means we'd need to further dirty the `openstack` resource `Backup` to exclude BMHs and their secrets/CMs -- the latter requiring new labeling code, perhaps, in the OpenStack Baremetal operator.

backup-restore/restore/README.md
```diff
@@ -170,6 +173,16 @@ data:
       - patchData: |
           spec:
             disabled: "True"
+    # BaremetalHosts — pause BMHs during restore so Metal3 does not act
+    # on them before their status is restored. Uses JSON patch because
+    # the annotation key contains "/".
+    - conditions:
+        groupResource: baremetalhosts.metal3.io
+        resourceNameRegex: ".*"
```
stuggi: is this required to add the resourceNameRegex? iiuc when omitted  the rule applies to all resource matching the conditions, in our case groupResource

backup-restore/restore/README.md
```diff
@@ -170,6 +173,16 @@ data:
...
+      patches:
```
stuggi: maybe we should also move to use mergePatches as I did, just to be consistent? the / did work for me using mergePatches.

backup-restore/restore/README.md
abays: Good point

backup-restore/restore/README.md
abays: Yes, will change to keep things consistent

stuggi (approved):
/lgtm

PR #176, opened 2026-05-12
[b/r] Add Swift xattr pre-backup hook to PVC backup

stuggi:
Add a Velero pre-backup hook to the PVC Backup CR (Step 4) that dumps Swift extended attributes (xattrs) before the CSI snapshot. OADP DataMover uses kopia which does not preserve xattrs during filesystem-level backup. Swift stores object metadata in xattrs, so without this hook all objects become inaccessible after restore.

The hook targets Swift storage pods by label and is silently skipped if Swift is not deployed. On the restore side, an init container in the swift-operator restores xattrs from the dump before Swift services start.

Also adds a note to the design doc's swift-operator PVC row explaining the xattr backup.

Jira: OSPRH-30020
Related: OSPRH-29819


stuggi:
depends-on: https://github.com/openstack-k8s-operators/swift-operator/pull/489

fmount:
+1, let's wait for the swift patch first and any potential update to this doc before landing.

fmount (approved):

PR #177, opened 2026-05-20
[backup-restore] Update galera restore docs for shortened resource names

stuggi:
Update the user guide and manual database restore docs to reflect the shortened resource names from mariadb-operator. GaleraRestore CR names no longer need a redundant "restore" suffix, and restore pod names follow the pattern restore-<cr-name> instead of <galera>-restore-<cr-name>. Document the galerarestore/name label for discovering restore pods.

Jira: OSPRH-30196
Depends-On: https://github.com/openstack-k8s-operators/mariadb-operator/pull/463


abays (approved):
/lgtm

PR #178, opened 2026-06-11
Update the AppCred section with FR6 changes

Deydra71:
Updated doc to reflect FR6 changes:
- Immutable Secrets with unique names, consumer finalizer handshake, Keystone AC revocation, EDPM NodeSet hash sync
- Removed stale `Future Improvements` section (all items now implemented) and old limitations (mutable secrets, no service-side finalizers, no revocation)
- Corrected AC CR deletion behavior: no password fallback, consumer finalizer keeps old Secret available, Keystone-side revocation is independent
- And other minor updates

### DIFF
--- a/application_credentials.md
+++ b/application_credentials.md
@@ -18,30 +18,34 @@ Three layers of operators are involved:
 ┌─────────────────────────────────────────────────────┐
 │  openstack-operator                                 │
 │  - Owns KeystoneApplicationCredential CRs           │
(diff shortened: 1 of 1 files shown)


fmount (approved):
/lgtm thanks for this patch. It is in line with the current implementation/approach and this is extremely important as `dev-docs` is becoming the source of truth for AI assistants.

PR #179, opened 2026-06-18
Update baremetal B/R to use skip/pause annotations

abays:
We have added webhook-skip and reconciliation-pause annotations to the OpenStack Baremetal operator.  These annotations allow for us to restore `OpenStackBaremetalSets` without manually disabling the operator's webhooks and scaling down its replica count to 0.  So we need to update the associated backup and restore docs to reflect this improved usage.

Jira: https://redhat.atlassian.net/browse/OSPRH-30426

### DIFF
--- a/backup-restore/restore/README.md
+++ b/backup-restore/restore/README.md
@@ -19,7 +19,7 @@ before starting the next.
 | 30 | OpenStackControlPlane, Reservation | **Restored with `deployment-stage: infrastructure-only` annotation** |
 | 40 | GaleraBackup, IPSet, DataPlaneService | Backup config, IP sets, custom DataPlane services |
 | 50 | **Manual**: Database restore | Create GaleraRestore CRs, restore databases, remove deployment-stage annotation |
-| 55 | BaremetalHosts (Step 7), OpenStackBaremetalSet (Step 8) | Optional: baremetal-provisioned nodes only. BMH secrets/CMs restored here if separate namespace. Requires BMH pause annotation (resource modifier), operator scale-down, and `restoreStatus` |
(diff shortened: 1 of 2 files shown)


backup-restore/restore/README.md
```diff
@@ -182,6 +182,21 @@ data:
...
               baremetalhost.metal3.io/paused: ""
+    # OpenStackBaremetalSets — skip webhook validation and pause
+    # reconciliation during restore. Velero restores spec and status in
+    # separate API calls; the webhook rejects the restore because the
+    # status is initially empty, and the reconciler would act on
+    # incomplete state before the status is set.
+    - conditions:
+        groupResource: openstackbaremetalsets.baremetal.openstack.org
+        namespaces:
+        - openstack
```
stuggi: just realized that we have namespaces in the rules. I think we could just remove them as we want them to happen to all objects get restored from the backup independent from the namespace? and remove them from all rules above, too

backup-restore/user-guide.md
```diff
@@ -450,6 +450,21 @@ data:
...
               baremetalhost.metal3.io/paused: ""
+    # Rule 5: OpenStackBaremetalSets — skip webhook validation and pause
+    # reconciliation during restore. Velero restores spec and status in
+    # separate API calls; the webhook rejects the restore because the
+    # status is initially empty, and the reconciler would act on
+    # incomplete state before the status is set.
+    - conditions:
+        groupResource: openstackbaremetalsets.baremetal.openstack.org
+        namespaces:
+        - openstack
```
stuggi: same as above

backup-restore/restore/README.md
abays: Sure, should I update those other ones now or do a follow-up to keep this cleaner?

stuggi: I think its good to do it now

abays: Done

backup-restore/user-guide.md
abays: Done

stuggi (approved):
/lgtm

PR #180, opened 2026-06-19
[backup-restore] Remove redundant includedNamespaces from PVC backup hook

stuggi:
The `includedNamespaces` on the swift-xattr-backup hook is redundant because the Backup CR already scopes to the `openstack` namespace via `spec.includedNamespaces`. Hooks only target pods within the backup's namespace scope regardless.


abays (approved):
/lgtm

PR #185, opened 2026-09-16
Add how to test code with nfs in multinode envs

elvgarrui:
The guide previously featured a more CRC focused approach. Added info about multinode environments as it could be useful for many situations where CRC could not be enough.


averdagu:
looks good to me!

stuggi (approved):
/lgtm

PR #186, opened 2026-09-17
Document gated OLM upgrade workflow with install_yamls

stuggi:
Add an install_yamls option to the "Deploying the Update" chapter showing how to gate an operator upgrade: install a pinned starting CSV first with manual InstallPlan approval, then approve each version via the openstack_approve_installplan target.


abays (approved):
/lgtm

